Issue #15, opened 2024-09-22
Upgrade Bun to newer version that supports windows


YoavCodes:
Now that Electrobun is more stable we can finally update the version of Bun we use to a more recent one that has windows support.

We've been using a really old version of bun this whole time while focusing on Electrobun stability.

The biggest API change has probably just been reading/writing from named pipes and streams

This will unblock the upcoming cross-platform work

PR #16, opened 2024-09-22
Upgrade bun to a version that supports Windows OS

YoavCodes:
Fixes https://github.com/blackboardsh/electrobun/issues/15


VityaSchel:
Can I safely download this patch and apply to my local electronbun to get windows exports?

YoavCodes:
> Can I safely download this patch and apply to my local electronbun to get windows exports?

unfortunately no.

This PR is just bumping the bun version (and fixes some stuff from bumping the zig version in a previous PR).

This new version of bun is after their 1.1 which supports Bun running on windows at all.

But actually making electrobun work on windows will still require me to:
- figure out windows application bundle format
- implement a windows installler and update electrobun's auot-update mechanism to work there.
- make and distribute windows builds of all the electrobun binaries (launcher, self-extractor, cli, etc.)
- most importantly I'll need to go through the native api wrapper that handles things like system tray, opening application windows and so on, and writing a native wrapper around windows apis. 
    - The current one in electrobun is written in objc that uses macos apis and the mac system webview (webkit)
    - I believe on windows I'll need to write one in c or c++ that uses windows system webview (edge)
    
So beyond this PR there's still quite a lot of work to get to windows support.

VityaSchel:
I believe in you and Electrobun, please keep doing great work!

PR #34, opened 2024-11-22
Capitalize "typescript"

wish13yt:
Noticed that in the "Write typescript for the main process and webviews without having to think about it. One language, no hassle." part, Typescript wasn't capitalized


YoavCodes:
Thanks!

PR #35, opened 2024-11-24
Fix docs typo

nadavspi-so:


YoavCodes:
already fixed in https://github.com/blackboardsh/electrobun/pull/31/files

PR #56, opened 2025-02-13
Add draggable region to element via attribute

NetOpWibby:
Just adds an option for folks who don't want to use a class to enable draggable functionality.


YoavCodes:
Thanks for the suggestion. but with an attribute as you've implemented then this becomes a draggable element

```
<div electrobun-webkit-app-region-drag="false"></div>
```

if you fix it to only be draggable when it's true then setting it to false is meaningless.

The original behaviour is a CSS property

```
-webkit-app-region: drag
```

and so a css class is the closest approximation given that electrobun is webview agnostic (we currently support CEF on all platforms and gtkwebkit, wkwebkit, and webview2).

so I'm closing this.

PR #58, opened 2025-03-06
Windows dev builds

YoavCodes:
## This PR
- Dev Electrobun apps on Windows
- Refactor zig layer, FFI, and RPC to use Bun FFI directly (Refactor most of Electrobun's core)


YoavCodes:
omfg I'm so glad that's over. onto Linux

PR #76, opened 2025-08-15
fix(cli,updater): resolve Windows tar extraction failures in MSYS/Git Bash

AugusDogus:
- Use relative paths instead of absolute paths when calling tar on Windows
- Prevents MSYS from interpreting drive letters (C:) as network hostnames
- Fixes "Cannot connect to C: resolve failed" errors
- Updates tar execSync calls in CLI, Updater, and build modules
- Maintains compatibility across all Windows shell environments

Affects:
- src/cli/index.ts: Updated 2 tar extraction locations
- src/bun/core/Updater.ts: Updated 1 tar extraction location
- build.ts: Updated 1 Windows CEF extraction location

[Relevant Discord thread](https://discord.com/channels/1190061229914464397/1405359210304241667).

### DIFF
--- a/build.ts
+++ b/build.ts
@@ -2,7 +2,7 @@
 
 import { $ } from "bun";
 import { platform, arch } from "os";
-import { join, dirname } from 'path';
+import { join, dirname, relative } from 'path';
 import { existsSync, readdirSync, renameSync, readFileSync, writeFileSync, mkdirSync } from "fs";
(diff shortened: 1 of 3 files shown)


AugusDogus:
I tested this solution with my previously failing `zsh` shell and with my previously working `cmd` shell. This solution works for both.

PR #83, opened 2025-09-03
fix(win): resolve WebView2 window resize deadlock on worker threads

AugusDogus:
## Problem
WebView2 window resize operations were freezing the application when called from worker threads. The issue occurred because `WebView2View::resize()` was using `dispatch_sync()` which blocked worker threads while waiting for the main thread to complete the resize operation. This created deadlocks due to conflicting thread synchronization between our dispatcher and WebView2's internal COM threading model.

## Solution
- Added `dispatch_async()` method to `MainThreadDispatcher` for non-blocking main thread dispatch
- Modified `WebView2View::resize()` to use async dispatch instead of sync dispatch
- Created separate message types to distinguish between synchronous and asynchronous operations

## Implementation Details
`WM_EXECUTE_ASYNC` was chosen for the async message type because:
- The calling thread continues immediately without blocking
- The task executes asynchronously on the main thread
- No response or completion notification is returned to the caller
This contrasts with `WM_EXECUTE_SYNC_BLOCK` which blocks the calling thread and returns a response.

## Why Async is Required
WebView2's `put_Bounds()` method must execute on the main UI thread, but calling it synchronously from worker threads creates deadlocks. The async approach ensures:
- Worker threads continue processing without blocking
- Resize operations complete reliably on the main thread
- No risk of thread synchronization conflicts
- Proper separation of concerns between UI and worker operations

## Manual Testing
- Window resize operations now complete without freezing
- Worker threads remain responsive during resize operations
- No performance impact on resize functionality

## Before PR - Failing
https://github.com/user-attachments/assets/97dd649e-db8a-402d-b278-3bef59c36c57

## After PR - Success
https://github.com/user-attachments/assets/faad2aa9-d60b-4fb5-baf7-0b6ee1574750

### DIFF
--- a/src/native/win/nativeWrapper.cpp
+++ b/src/native/win/nativeWrapper.cpp
@@ -71,6 +71,7 @@ using namespace Microsoft::WRL;
 // Ensure the exported functions have appropriate visibility
 #define ELECTROBUN_EXPORT __declspec(dllexport)
 #define WM_EXECUTE_SYNC_BLOCK (WM_USER + 1)
+#define WM_EXECUTE_ASYNC_BLOCK (WM_USER + 2)
 
 // Forward declarations
 class AbstractView;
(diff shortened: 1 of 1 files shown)


YoavCodes:
This is a top-notch PR. Thank you for finding, fixing, and then doing such a great writeup. Works great!

YoavCodes:
If I pushed the buttons right this PR should be live in `0.1.2`

PR #105, opened 2025-12-04
Cleaning up the logo

eduplessis:
Simplifying the logo


eduplessis:
Sorry for making changes to the logo without explanation.

The first thing that bothered me was that it was obvious it had been generated by AI. And because of that, the file and the quality of the lines in the SVG were really very poor.

When designing a logo, it's important to keep everything as simple as possible. That's why I removed some of the smaller stars, which only disappeared when the logo was reduced in size. The same goes for the nose. And as for the foot, let's just say it was distracting. You should always think of the logo as if it were printed in a 1cm x 1cm format or embroidered really small and still need to be recognizable.

For the rabbit's front legs, I reduced the dividing line that went up to the throat.

I redid the logo while trying to stay true to the original. If you want a more comprehensive and complete approach, I can of course do

eduplessis:
Oh and how i cleaned up... I've remade it by hand with affinity studio.

MulverineX:
What would be the context in which the electrobun logo would actually be that small? As far as I can tell there isn't really a use-case for it being a favicon, an app icon, nor anything else equivalent. 

I do agree that the vector's finer details should be touched up, but removing the smaller stars & making it mono-tone really doesn't make much sense IMO.

YoavCodes:
closing this. 

if anyone else tries to mess with the logo **_I'll send my sausage dog after you._**

PR #106, opened 2025-12-06
[PATCH] Fix Dev Setup and Building on Linux - v2

NullDev:
Fixed new bash setup script to correctly resolve paths & tiny bit of cleanup and typization 

Edit: No clue why the bun lock shows up as deleted. It shouldn't be 
image

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -15,6 +15,7 @@ import {
   copyFileSync,
   renameSync,
 } from "fs";
+import type { Subprocess } from "bun";
 import { execSync } from "child_process";
 import * as readline from "readline";
 import tar from "tar";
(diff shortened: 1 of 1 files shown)


YoavCodes (changes requested):
I think i need to bring back the launcher on linux. it’s probably going to be a bit involved with CI and such so might be something i have to tackle.

I’m not sure your changes fix the issue because desktop shortcut is still not created.

Development shouldn’t be blocked though because the cli still loads the app in dev mode as far as i can tell (with and without your pr).

let me know if I missed something

NullDev:
I only tested dev mode with `dev:playground:clean:linux`
Without this PR, that command does not work for me.
Did not test prod / builds but I can look into that

PR #108, opened 2025-12-09
feat: Add Always on Top window functionality

J3n5en:
## Summary
  - Add `alwaysOnTop` option to `BrowserWindow` constructor
  - Add `setAlwaysOnTop(boolean)` method for runtime toggling
  - Add `alwaysOnTop` getter to check current state

  ## Changes
  - **macOS**: Use `NSFloatingWindowLevel` / `NSNormalWindowLevel`
  - **Linux**: Use `gtk_window_set_keep_above` (GTK) or
  `_NET_WM_STATE_ABOVE` (X11)
  - **Windows**: Use `SetWindowPos` with `HWND_TOPMOST` /
  `HWND_NOTOPMOST`

  ## Usage
  ```typescript
  // Set on creation
    alwaysOnTop: true,
    // ...
  });

  // Toggle at runtime
  win.setAlwaysOnTop(true);
  win.setAlwaysOnTop(false);

  // Check state
  console.log(win.alwaysOnTop);

```

Test Plan

  - Run bun run:playground and test Always on Top checkbox in Window
  Demo
  - Run bun test and execute the "Window Always on Top" test case
  - Verify window stays on top when other windows are focused
  - Verify toggling off allows window to go behind others

### DIFF
--- a/src/bun/core/BrowserWindow.ts
+++ b/src/bun/core/BrowserWindow.ts
@@ -71,7 +71,7 @@ export class BrowserWindow<T> {
   };
   // todo (yoav): make this an array of ids or something
   webviewId: number;
-  alwaysOnTop: boolean = false;
+  private _alwaysOnTop: boolean = false;
 
   constructor(options: Partial<WindowOptionsType<T>> = defaultOptions) {
     this.title = options.title || "New Window";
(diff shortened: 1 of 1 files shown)


YoavCodes:
sorry I've been heads down getting to v1 (almost there). closing this because always on top is already implemented in the next stable release and works across all OSes.

PR #119, opened 2026-01-20
Fix WKWebView crash on tab close

Acumen-Desktop:
### Problem
When closing a WKWebView tab, the app crashes with signal 11 (SIGSEGV).

### Root Cause
1. **Dangling pointer**: The [remove](cci:1://file:///Users/james/Documents/Software/Blackboard/colab-clone/src/main/index.ts:1380:8-1382:9) method was releasing the webview and then accessing it in an async `dispatch_async` block, causing a use-after-free
2. **KVO observer leak**: The fullscreenState KVO observer was only removed in `dealloc`, but by then `self.webView` was already nil

### Solution
- Remove KVO observer first, before any cleanup
- Remove webview from superview synchronously instead of async dispatch
- Add safety checks in `dealloc` for already-removed observers

### Testing
Tested by repeatedly opening and closing both CEF and WebKit tabs - no more crashes.

### DIFF
--- a/package/src/native/macos/nativeWrapper.mm
+++ b/package/src/native/macos/nativeWrapper.mm
@@ -4,38 +4,38 @@
  * =============================================================================
  */
 
-#import <WebKit/WebKit.h>
-#import <objc/runtime.h>
 #import <Cocoa/Cocoa.h>
-#import <Foundation/Foundation.h>
 #import <CommonCrypto/CommonCrypto.h>
+#import <Foundation/Foundation.h>
 #import <QuartzCore/QuartzCore.h>
 #import <UserNotifications/UserNotifications.h>
+#import <WebKit/WebKit.h>
+#import <objc/runtime.h>
 
 // CEF includes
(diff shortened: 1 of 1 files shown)


Acumen-Desktop:
Update: Tested this PR on my Intel Mac (i7-7700K, macOS 13.7.8 Ventura). After building from source with bun build.ts, the app runs successfully. The published npm binaries still crash with WKWebsiteDataStore dataStoreForIdentifier: unrecognized selector since that API is macOS 14+ only. The @available(macOS 14.0, *) check in the source works correctly when built locally.

YoavCodes:
Screenshot 2026-02-04 at 4 57 10 PM

Thanks for taking the time, but there is no universe in which I'm going to code review a +6,000 line PR.

I don't even know what version of electrobun you experienced the crash in. next time please submit a proper github issue with all the details and repro steps instead of randomly re-writing the entire macos implementation on a whim.

electrobun doesn't even have tabs so I have no idea what this is even trying to fix.

Issue #93, opened 2025-11-02
Allow macOS traffic light button repositioning


florinsdistortedvision:
Electron allows hiddenInset like Electrobun currently does, but also allows manually setting the X and Y offset for the native traffic light buttons.

```js
const { BrowserWindow } = require('electron')

const win = new BrowserWindow({
  titleBarStyle: 'hidden',
  trafficLightPosition: { x: 10, y: 10 }
})
```

Taken from: https://www.electronjs.org/docs/latest/tutorial/custom-title-bar#customize-the-traffic-light-position-macos

YoavCodes:
instead of this i've enabled frameless and transparent windows and exposed all the window control functions, so you can build traffic lights in html/css and put them wherever you want. try the latest betas and see the kitchen sink for api examples until the new docs are up

PR #131, opened 2026-01-30
Add windowButtonPosition to title bar config for hiddenInset title bar styles

rjschie:
### Motivation

This allows us to position the window buttons located in in the title bar for added customizability.
Closes: https://github.com/blackboardsh/electrobun/issues/93

Added bonus: this is a more complete solution than Electron (and maybe Tauri) is using since they have a ghost hover activation bug when you hover over the original traffic light positions

Here's a summary from Claude on the button repositioning with swizzle and why we needed it:

>   Button positioning
> 
>   `applyWindowButtonPosition` is called on window show and resize. It creates a `ButtonsAreaHoverView` overlay in the
>   `NSTitlebarContainerView` and calls `redraw`, which:
> 
>   1. Resizes `NSTitlebarContainerView` to fit the new margin + button height, pinned to the top of the window
>   2. Positions the three standard buttons (close/minimize/zoom) at Y=0 within the container using `setFrameOrigin`
>   3. Places a transparent `ButtonsAreaHoverView` overlay over the buttons to provide correct hover tracking (since it manages its
>   own `NSTrackingArea`)
> 
>   Ghost hover fix: `_mouseInGroup:` swizzle
> 
>   After repositioning, hovering over the original button locations would activate the repositioned button's hover states. This happens because `NSThemeFrame` (the window's
>    root frame view) has internal tracking areas that call the private `_mouseInGroup:` method to determine whether the mouse is over
>    the traffic light button group. These tracking areas use the original button positions and can't be removed without breaking
>   other window behavior (`NSThemeFrame` has ~10 tracking areas for various purposes).
> 
>   The fix swizzles `_mouseInGroup:` on `NSThemeFrame` using `method_setImplementation`. The replacement checks whether the mouse
>   location falls within the repositioned button group rect (converted to `NSThemeFrame` coordinates) instead of relying on the
>   default hit-testing logic. The original implementation is preserved and used as a fallback when buttons haven't been
>   repositioned.

### Screenshots
Default titleBarStyle "hiddenInset" on left, titleBarStyle "hiddenInset" with custom windowButtonPosition on right (x: 15, y: 20)
windowButtonPosition

Ghost Hover issue compared to Electron

https://github.com/user-attachments/assets/a07902b5-4358-4d67-b649-848083ed2d4b

### Considerations
-  Should I add a method `getWindowButtonPosition` so we can ensure the automated tests validate the actual position? Otherwise, I couldn't think of a use case to know the position at runtime, but if you think it would be valuable I can add that.
- Let me know if you want to rename it to match what others do (e.g. `trafficLightPosition`)
- Right now the x & y values set the absolute position from top left (0, 0) so a value of (x: 0, y: 0) would put the buttons all the way into the corner. However, I can update this to be an offset from the default position if you think that would be more valuable. I figured it would help users define their layout if the controlled it the same as the HTML page (knowing that it starts at [0, 0] from the upper left).

### DIFF
--- a/package/src/native/macos/nativeWrapper.mm
+++ b/package/src/native/macos/nativeWrapper.mm
@@ -6,6 +6,7 @@
 
 #import <WebKit/WebKit.h>
 #import <objc/runtime.h>
+#import <objc/message.h>
 #import <Cocoa/Cocoa.h>
 #import <Foundation/Foundation.h>
 #import <CommonCrypto/CommonCrypto.h>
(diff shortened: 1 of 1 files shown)


YoavCodes:
Thanks for taking the time! But i'm not a fan of this solution because it:
- requires always sending the offset over ffi
- is very mac specific
- has lots of custom logic for positioning the buttons in nativeWrapper and swizzling which is super brittle across macos updates and versions.


**Instead**
I've added some new functionality in v1 that'll be in the next stable build which lets you:
- create frameless windows where you can define your own window chrome entirely in html/css/js
- exposed window minimize/maximize/close etc. to the bun api which you can add to your browser rpc and call from the browser
- works cross platform

PR #134, opened 2026-02-04
fix(windows): resolve dev command ENOENT error on Windows

### LINKED ISSUE #133: Windows: `electrobun dev` fails with ENOENT on `./bun.exe` due to missing `else` block in compiled CLI
# Windows: `electrobun dev` fails with ENOENT on `./bun.exe` due to missing `else` block in compiled CLI

## Summary

The Windows CLI binary has a compilation bug where the `else` block between two `Bun.spawn()` calls is stripped during the build process. This causes both spawn attempts to execute sequentially instead of being mutually exclusive, resulting in a fatal `ENOENT` error.

## Affected Versions

- ✅ Confirmed in **v0.8.0** (stable)
- ✅ Confirmed in **v1.11.0-beta.0** (latest beta)
- Likely affects **all Windows CLI releases**

## Environment

- **OS**: Windows 11 (win32 x64)
- **Bun**: v1.3.8
- **Node**: v23.9.0

## Steps to Reproduce

1. Create a new electrobun project on Windows (or clone an existing one)
2. Install dependencies: `bun install`
3. Run: `bun run electrobun dev`

## Expected Behavior

The CLI should check if `launcher.exe` exists:
- If yes → spawn `launcher.exe`
- If no → spawn `./bun.exe` as fallback

Only ONE of these should execute.

## Actual Behavior

Both spawn calls execute sequentially. The first (launcher.exe) succeeds, but then the second (./bun.exe) also runs and fails with:

```
Fatal error: 40555 |       if (existsSync(launcherPath)) {
40556 |         mainProc = Bun.spawn([launcherPath], {
40557 |           stdio: ["inherit", "inherit", "inherit"],
40558 |           cwd: bundleExecPath
40559 |         });
40560 |         mainProc = Bun.spawn(["./bun.exe", "../Resources/main.js"], {
                                    ^
ENOENT: no such file or directory, uv_spawn './bun.exe'
    path: "./bun.exe",
 syscall: "uv_spawn",
   errno: -4058,
    code: "ENOENT"

      at main (B:/~BUN/root/electrobun:40560:29)
```

### Source Code (Correct)

In `src/cli/index.ts` lines 2221-2238, the logic is correct with proper `if-else`:

```typescript
} else if (OS === 'win') {
  // Windows: Use launcher if available, otherwise fallback to direct execution
  const launcherPath = join(bundleExecPath, 'launcher.exe');
  if (existsSync(launcherPath)) {
    mainProc = Bun.spawn([launcherPath], {
      stdio: ['inherit', 'inherit', 'inherit'],
      cwd: bundleExecPath
    })
  } else {
    // Fallback for older builds
    mainProc = Bun.spawn(['./bun.exe', '../Resources/main.js'], {
      stdio: ['inherit', 'inherit', 'inherit'],
      cwd: bundleExecPath,
      onExit: (proc, exitCode, signalCode, error) => {
        console.log('Bun process exited:', { exitCode, signalCode, error });
      }
    })
  }
}
```

### Compiled Binary (Broken)

In the compiled `electrobun.exe`, the error trace shows lines 40559-40560:

```
40559 |         });
40560 |         mainProc = Bun.spawn(["./bun.exe", "../Resources/main.js"], {
```

Notice that line 40559 ends the first spawn with `});` and line 40560 **immediately** starts the second spawn. The `} else {` block is completely missing.

The compiled code effectively becomes:

```typescript
if (existsSync(launcherPath)) {
  mainProc = Bun.spawn([launcherPath], { ... });
}
// Missing: } else {
mainProc = Bun.spawn(["./bun.exe", ...], { ... });  // ALWAYS RUNS!
```

### Proof That launcher.exe Works

Running the launcher directly succeeds:

```bash
cd build/dev-win-x64/MyApp-dev/bin
./launcher.exe
# App launches successfully!
```

This confirms the issue is not with the launcher or the build, but with the CLI's spawn logic.

## Version Comparison

| Version | Error Line | Status |
|---------|-----------|--------|
| v0.8.0 | 40560 | ❌ Bug present |
| v1.11.0-beta.0 | 38844 | ❌ Bug present |

The line numbers differ slightly between versions, but the same bug exists in both.

## Likely Cause

This appears to be a **Bun compiler bug** when cross-compiling for Windows using `bun build --compile --target=bun-windows-x64`. During bundling/minification, the `else` keyword is being incorrectly stripped.

Alternatively, it could be an issue with:
- Dead code elimination incorrectly removing the `else`
- Minification/optimization passes
- The specific Bun version used in the build pipeline

### Option 1: Refactor to avoid the issue

Restructure the code to not rely on `else`:

```typescript
} else if (OS === 'win') {
  const launcherPath = join(bundleExecPath, 'launcher.exe');
  if (existsSync(launcherPath)) {
    mainProc = Bun.spawn([launcherPath], {
      stdio: ['inherit', 'inherit', 'inherit'],
      cwd: bundleExecPath
    });
  }

  if (!existsSync(launcherPath)) {  // Explicit second check instead of else
    mainProc = Bun.spawn(['./bun.exe', '../Resources/main.js'], {
      stdio: ['inherit', 'inherit', 'inherit'],
      cwd: bundleExecPath
    });
  }
}
```

### Option 2: Use early return pattern

```typescript
} else if (OS === 'win') {
  const launcherPath = join(bundleExecPath, 'launcher.exe');
  if (existsSync(launcherPath)) {
    mainProc = Bun.spawn([launcherPath], {
      stdio: ['inherit', 'inherit', 'inherit'],
      cwd: bundleExecPath
    });
    // Continue to process.on("SIGINT") setup below
  } else {
    mainProc = Bun.spawn([join(bundleExecPath, 'bun.exe'), join(bundleResourcesPath, 'main.js')], {
      stdio: ['inherit', 'inherit', 'inherit'],
      cwd: bundleExecPath
    });
  }
}
```

Also note: the fallback uses `'./bun.exe'` which is a Unix-style relative path. On Windows, this should probably be `join(bundleExecPath, 'bun.exe')` for reliability.

### Option 3: Report to Bun

If this is a Bun compiler bug, it should be reported to https://github.com/oven-sh/bun/issues with a minimal reproduction case.

## Workaround

Until this is fixed, users can create a wrapper script that:
1. Runs `electrobun build` (which works correctly)
2. Manually spawns `launcher.exe` directly

Example (`scripts/dev.cjs`):

```javascript
const { spawn, spawnSync } = require('child_process');
const { existsSync, readdirSync } = require('fs');
const { join } = require('path');

const isWindows = process.platform === 'win32';

if (isWindows) {
  // Run build
  spawnSync('bun', ['x', 'electrobun', 'build'], { stdio: 'inherit', shell: true });

  // Find and launch app directly
  const appDir = /* find build directory */;
  const launcherPath = join(appDir, 'bin', 'launcher.exe');
  spawn(launcherPath, [], { stdio: 'inherit', cwd: join(appDir, 'bin') });
} else {
  // Linux/macOS work correctly
  spawn('bun', ['x', 'electrobun', 'dev'], { stdio: 'inherit', shell: true });
}
```

## Additional Notes

1. The `electrobun build` command works correctly - only `electrobun dev` is affected
2. The launcher.exe itself works perfectly when invoked directly
3. This bug blocks all Windows development workflows using `electrobun dev`

## Related Issues

- #73 - Windows build error (different issue - C++ compiler setup)
- #100 - Windows native tar freezes (different issue)
- #39 - Windows Dev Builds (ARM discussion)

None of the existing issues cover this specific bug.

YoussefZayed:
## Summary

Fixes #133 - Windows `electrobun dev` fails with `ENOENT` on `./bun.exe`

The Windows CLI binary was failing because both `Bun.spawn()` calls were executing sequentially instead of being mutually exclusive.

## Root Cause

During Bun's `--compile` process for Windows targets, the `else` block between the two spawn calls was being stripped. This caused the compiled binary to effectively become:

```typescript
if (existsSync(launcherPath)) {
  mainProc = Bun.spawn([launcherPath], { ... });
}
// else block stripped!
mainProc = Bun.spawn(["./bun.exe", ...], { ... });  // Always runs!
```

## Changes

- Store `existsSync()` result in a variable to avoid the if-else pattern
- Use explicit `if (!launcherExists)` instead of `else` block to work around the Bun compiler bug
- Fix Unix-style relative paths (`./bun.exe`, `../Resources/main.js`) to use proper `join()` calls with `bundleExecPath` and `bundleResourcesPath`

## Testing

Before this fix:
```
ENOENT: no such file or directory, uv_spawn './bun.exe'
```

After this fix: The dev command should correctly spawn only `launcher.exe` when it exists.

## Notes

This is a workaround for what appears to be a Bun compiler bug when cross-compiling for Windows. A separate issue may need to be filed with Bun if this pattern affects other codebases.

### DIFF
--- a/package/src/cli/index.ts
+++ b/package/src/cli/index.ts
@@ -3926,14 +3926,22 @@ ${schemesXml}
 		} else if (OS === "win") {
 			// Windows: Use launcher if available, otherwise fallback to direct execution
 			const launcherPath = join(bundleExecPath, "launcher.exe");
-			if (existsSync(launcherPath)) {
+			const launcherExists = existsSync(launcherPath);
+
+			if (launcherExists) {
 				mainProc = Bun.spawn([launcherPath], {
 					stdio: ["inherit", "inherit", "inherit"],
 					cwd: bundleExecPath,
 				});
-			} else {
-				// Fallback for older builds
-				mainProc = Bun.spawn(["./bun.exe", "../Resources/main.js"], {
+			}
+
+			// Fallback for older builds without launcher
+			// Note: Using explicit !launcherExists check instead of else block
+			// to work around Bun compiler bug that strips else blocks on Windows
+			if (!launcherExists) {
+				const bunExePath = join(bundleExecPath, "bun.exe");
+				const mainJsPath = join(_bundleResourcesPath, "main.js");
+				mainProc = Bun.spawn([bunExePath, mainJsPath], {
 					stdio: ["inherit", "inherit", "inherit"],
 					cwd: bundleExecPath,
 					onExit: (_proc, exitCode, signalCode, error) => {


YoussefZayed:
## Testing Notes

I attempted to build and test the fix locally on Windows but encountered build dependencies issues (cmake required). Here's what I was able to verify:

### Minimal Compilation Test
Created isolated test cases with the same if-else pattern and compiled them with `bun build --compile --target=bun-windows-x64-baseline`:

```typescript
// Original pattern
if (existsSync(launcherPath)) {
  mainProc = { cmd: [launcherPath] };
} else {
  mainProc = { cmd: ["./bun.exe", ...] };
}

// Fixed pattern  
const launcherExists = existsSync(launcherPath);
if (launcherExists) {
  mainProc = { cmd: [launcherPath] };
}
if (!launcherExists) {
  mainProc = { cmd: [bunExePath, mainJsPath] };
}
```

Both patterns produced identical correct behavior in isolated tests. The bug may be specific to the larger bundled codebase or specific bundling conditions in CI.

### What the Fix Does
1. **Stores `existsSync()` result in a variable** - Avoids potential issues with the if-else pattern
2. **Uses explicit `if (!launcherExists)` check** - Guarantees mutual exclusion even if else gets stripped
3. **Fixes Unix-style paths** - Changed `"./bun.exe"` and `"../Resources/main.js"` to use proper `join()` calls

### Recommendation
The electrobun CI should build and test this on Windows to verify the fix resolves the issue in the full bundled binary. The fix is logically equivalent to the original and should not introduce any regressions.

YoussefZayed:
## ✅ Fix Verified - Successfully Tested on Windows

I was able to compile and test the fix. Here are the results:

### Test Environment
- Windows 11 (win32 x64)
- Bun v1.3.8
- electrobun v1.11.0-beta.0

### Build Process
1. Cloned electrobun repo
2. Created stub `templates/embedded.ts` (auto-generated in CI)
3. Compiled CLI: `bun build src/cli/index.ts --compile --target=bun-windows-x64-baseline --outfile bin/electrobun.exe`

### Test Results

**Before fix:**
```
ENOENT: no such file or directory, uv_spawn './bun.exe'
```

**After fix:**
```
Using config file: electrobun.config.ts
Launcher starting on windows...
Current directory: ...\LarrysLameMission-dev\bin
Spawning: ...\bin\bun.exe ...\Resources\main.js
Child process spawned with PID anyopaque@200
[LAUNCHER] Loading app code from flat files
```

The app launches correctly via `launcher.exe` without hitting the fallback path or any ENOENT errors.

### Note
Also fixed a variable name issue (`bundleResourcesPath` → `_bundleResourcesPath`) that was caught during testing.

YoavCodes:
This PR attempts to fix a bug report that can't be reproed in a way that extends dead code. I already removed the dead code on main so closing this.

Issue #158, opened 2026-02-18
Bug: Missing GTK/WebKit dependencies causes opaque launcher failure for Linux


bxrne:
# Problem

Tried the multi browser template project but `bun start` fails to launch the application due to missing runtime dependencies it links against.

It fails at startup due to missing system GTK/WebKit libraries. The error message is low-level (ERR_DLOPEN_FAILED) and does not indicate which packages need to be installed.

- These Linux runtime dependencies are not documented.
- The launcher surfaces a low-level dlopen error.
- There is no guidance for which packages to install. (`ldd libNativeWrapper.so` can show it to be fair)

## Output

```bash
Failed to open library ".../libNativeWrapper.so":
libwebkit2gtk-4.1.so.0: cannot open shared object file

code: ERR_DLOPEN_FAILED
```

# Environment

- OS: Arch Linux x86_64
- Kernel: Linux 6.18.1-arch1-Watanare-T2-1-t2
- WM: Hyprland 0.53.3 (Wayland)

YoavCodes:
Fixed in the next stable release.

PR #159, opened 2026-02-18
fix: #158 - Add missing linux dependency documentation and better DLOpen error messaging

bxrne:
## Summary
Fixes first-run failures on Arch Linux caused by missing GTK/WebKit runtime libraries by:
- Documenting Linux runtime dependencies in `README.md`.
- Printing a clearer launcher error when common Linux libraries are missing.

On Arch Linux, launching the multi-tab browser template failed with missing shared library errors. The launcher error was opaque (`ERR_DLOPEN_FAILED`). These dependencies are hard runtime requirements for the Linux native wrapper and should be documented and surfaced clearly at runtime.

## Changes
- Added a `Linux Runtime Dependencies` section to `README.md` with Arch and Ubuntu/Debian package commands.
- Updated the launcher to detect common missing libraries (`libwebkit2gtk-4.1`, `libgtk-3`, `libayatana-appindicator`) and print install instructions.

## Reproduction 
See #158 

## Related Issue
- #158

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,26 @@ Building apps with Electrobun is as easy as updating your package.json dependenc
 
 **This section is for building Electrobun from source locally in order to contribute fixes to it.**
 
+## Linux Runtime Dependencies
+
+Electrobun requires system GTK/WebKit libraries at runtime on Linux.
+
+### Arch Linux
+
+```bash
+sudo pacman -S webkit2gtk-4.1 gtk3 libayatana-appindicator
+```
+
+### Ubuntu/Debian
+
+```bash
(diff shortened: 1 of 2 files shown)


YoavCodes:
Thanks for taking the time, but the idea behind our official Ubuntu support policy is that:
- only use dependencies typically already installed on stock ubuntu
- if they're not then those dependencies should be auto-installed during dev/runtime and/or bundled into the app itself.

The idea of starting to maintain unsupported distro specific install commands or lists and have to manually keep updating those is not great.

If you can find a way to auto-generate that then I'd consider merging it, but I think the proper place for manual docs that hack electrobun to work on obscure distros is outside the repo

PR #169, opened 2026-02-20
chore: add .gitignore for templates

### LINKED ISSUE #162: Templates generated by `electrobun init` are missing .gitignore
## Summary

Projects scaffolded via `electrobun init` do not include a `.gitignore` file. This affects all 5 templates (hello-world, svelte, photo-booth, multitab-browser, react-tailwind-vite).

## Root Cause

In `package/build.ts`, the `generateTemplateEmbeddings()` function skips all dotfiles:

```typescript
entry.name.startsWith(".") ||  // All dotfiles skipped
```

This means even if a `.gitignore` is added to a template directory, it won't be embedded into the CLI.

## Impact

Users who run `electrobun init` and then `git init` will have no `.gitignore`, risking accidental commits of `node_modules/`, `build/`, `dist/`, etc.

## Suggested Fix

Either:
1. **Allowlist `.gitignore`** in the dotfile skip logic in `build.ts`, and add `.gitignore` files to each template
2. **Generate `.gitignore` in the CLI** during `electrobun init`, independent of templates

A typical `.gitignore` for Electrobun projects would include:
```
node_modules/
dist/
build/
*.tsbuildinfo
```

G3root:
fixes #162. add .gitignore files for the template.

### DIFF
--- a/templates/hello-world/.gitignore
+++ b/templates/hello-world/.gitignore
@@ -1,0 +1,25 @@
+# Dependencies
+node_modules/
+
+# Electrobun
+.electrobun/
+
+
+# Build output
+build/
+dist/
+*.app
+
+# OS
+.DS_Store
+Thumbs.db
+
+# Logs
+*.log
+npm-debug.log*
+
+# Environment
+.env
+.env.local
+.env.*.local
+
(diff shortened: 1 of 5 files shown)


YoavCodes:
I appreciate the attempt but closing this for a couple reasons.
- it only adds .gitignore for some of the templates not all of them
- ignoring `.electrobun/` 
- and the root issue is that the template embedding code actually skips any file that starts with a `.` as explained in the issue you linked to.

PR #170, opened 2026-02-20
chore: upgrade react template to tailwind v4

G3root:
upgrades react template to tailwind v4

### DIFF
--- a/templates/react-tailwind-vite/package.json
+++ b/templates/react-tailwind-vite/package.json
@@ -20,10 +20,8 @@
 		"@types/react": "^18.3.12",
 		"@types/react-dom": "^18.3.1",
 		"@vitejs/plugin-react": "^4.3.4",
-		"autoprefixer": "^10.4.20",
 		"concurrently": "^9.1.0",
-		"postcss": "^8.4.49",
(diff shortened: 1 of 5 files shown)


13RTK:
You did what I want to do😂

m1guer:
this version works?

YoavCodes (changes requested):
I'm not sure the templates need updating this often (they're just meant as proof of concept)

but my AI friend has requested some changes

```
I’d not merge #170 as-is.

  Main blocker:

  1. High: missing dependency for new plugin import.

  - PR changes vite.config.ts to import tailwindcss from "@tailwindcss/vite", but templates/react-tailwind-vite/
    package.json does not add @tailwindcss/vite.
  - Result: fresh template likely fails at runtime/build with module-not-found.

  Secondary:
  2. Low: template docs/lockfile drift.

  - It deletes tailwind.config.js/postcss.config.js, but template README still references Tailwind config.
  - bun.lock in template is not updated to reflect v4 stack.

  So the idea is fine, but it needs a follow-up commit before merge.
```

YoavCodes:

PR #172, opened 2026-02-20
fix monorepo-broken version detection and package resolution

AaronO:
`ELECTROBUN_DEP_PATH` was hardcoded to `projectRoot/node_modules/electrobun` which doesn't exist in monorepos with hoisted `node_modules`. This caused version detection to fall back to `"latest"` and all binary/cache paths & URLs to break.

- Walk up directory tree to find `electrobun` in `node_modules` (handles `yarn/pnpm/npm/bun` workspaces)
- Derive `.electrobun-cache` path from resolved location instead of hardcoding against `projectRoot`
- Replace runtime `package.json` reads with compile-time version constant


YoavCodes (approved):
Thanks!

PR #173, opened 2026-02-20
fix(build): Replace stuck tar with 7-Zip for Windows CEF extraction

adam2am:
### Issue
Hey, so when building Electrobun on Windows 10 Pro CEF extraction just... froze. tar.exe sat at 0% CPU, 0 progress for 10+ minutes. Had to kill it. No disk activity, just hangs. Basically made Electrobun unbuildable on Windows without manual intervention.

### Solution 
Use 7zip. Check if its installed.
- auto-install via winget if not present, fall back to tar if that fails. 
- Two-step extraction: .bz2 -> .tar -> contents

### Impact
- Before: tar hangs indefinitely, build fails
- After: 7-Zip extracts in 1-2 minutes, reliable
- Fallback: Still tries tar if 7-Zip unavailable

Tested on Windows 10 Pro, happy to hear your thoughts

### DIFF
--- a/package/build.ts
+++ b/package/build.ts
@@ -2,7 +2,7 @@
 
 import { $ } from "bun";
 import { platform, arch } from "os";
-import { join, dirname, relative } from "path";
+import { join, dirname, relative, basename } from "path";
 import {
 	existsSync,
 	readdirSync,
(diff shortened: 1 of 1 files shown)


justin-m-lacy:
I had a similar problem and it turned out to be a permissions issue with the VSCode terminal.
Running cmd terminal outside VSCode worked fine.
Not claiming your problem is the same, but it's possible.

YoavCodes:
> Windows 10 Pro

Sorry. We don't support Windows 10. It's been EOL for months already and will stop even getting extended security updates in October I think. 

I'd recommend you upgrade to Windows 11 or switch to Ubuntu, or MacOS as soon as possible.

PR #175, opened 2026-02-20
feat: add show option to BrowserWindow for deferred visibility

ragamufin:
## Summary

Adds a `show` option to `BrowserWindow` that controls whether the window is displayed immediately on creation. Defaults to `true` (current behavior).

When `show: false` is passed, the window is created but `showWindow()` is skipped. The window can be made visible later by calling `window.show()`.

## Motivation

Electron's `BrowserWindow` supports `show: false`, which is commonly used for:
- **E2E testing**: Playwright/test frameworks can spawn app instances that appear in the dock without stealing focus or flashing windows
- **Splash screens**: Create a window, load content, then show when ready
- **Background windows**: Create windows that start hidden until needed

This is a small, backwards-compatible change — existing code that doesn't pass `show` continues to work identically.

## Changes

- **`BrowserWindow.ts`**: Added `show?: boolean` to `WindowOptionsType`, added `show: true` to default options, passed through constructor → init → createWindow
- **`native.ts`**: Added `show?: boolean` to `createWindow` params, wrapped `showWindow()` in `if (params.show !== false)` guard

## Usage

```typescript
// Window appears immediately (default, unchanged behavior)
const win = new BrowserWindow({ title: "My App", url: "..." });

// Window created but not shown
const hidden = new BrowserWindow({ title: "Hidden", url: "...", show: false });

// Show it later
hidden.show();
```

## Testing

Tested locally on macOS ARM64:
- `new BrowserWindow()` with no `show` option → window appears immediately ✓
- `new BrowserWindow({ show: false })` → window created but not visible ✓
- `window.show()` on hidden window → window becomes visible ✓
- No crashes or errors ✓

### DIFF
--- a/package/src/bun/core/BrowserWindow.ts
+++ b/package/src/bun/core/BrowserWindow.ts
@@ -36,6 +36,8 @@ export type WindowOptionsType<T = undefined> = {
 	// Use for untrusted content (remote URLs) to prevent malicious sites from
 	// accessing internal APIs, creating OOPIFs, or communicating with Bun
 	sandbox: boolean;
+	/** Whether to show the window immediately on creation. Default: true */
+	show?: boolean;
 };
 
 const defaultOptions: WindowOptionsType = {
@@ -54,6 +56,7 @@ const defaultOptions: WindowOptionsType = {
 	transparent: false,
 	navigationRules: null,
 	sandbox: false,
+	show: true,
 };
 
 export const BrowserWindowMap: {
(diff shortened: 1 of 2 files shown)


a-eid:
any plans to merge this PR soon ? @YoavCodes ?

YoavCodes:
Appreciate the PR but this is sort of duplicate of the hidden property that landed recently.

https://github.com/blackboardsh/electrobun/blob/7fa590dc68ac23a19d936ec0960dbc827f461468/package/src/bun/core/BrowserWindow.ts#L35

PR #181, opened 2026-02-21
fix(windows): unique AppUserModelID for proper taskbar grouping

adam2am:
### Issue 

Hey, so windows taskbar creates duplicate icons for Electrobun apps:
1. Pin app to taskbar
2. Launch -> 2 separate icons appear (launcher.exe + bun.exe)
3. Pinned icon doesn't group with running process
4. Pinning bun.exe directly fails to launch app
**Root cause:** launcher.exe and bun.exe use different `AppUserModelID` values (or none), so Windows treats them as separate applications.

### Solution
Set unique `AppUserModelID` per app using `app.identifier` from config:
launcher.exe -> reads identifier -> sets AppUserModelID
 ->
passes via env var (ELECTROBUN_APP_IDENTIFIER)
 ->
bun.exe -> reads env var -> sets same AppUserModelID
Shortcuts also include `AppUserModelID` via COM `IPropertyStore` API.

### Core Fix
- `src/launcher/main.zig` - Read identifier, set AppUserModelID, pass to bun
- `src/native/win/nativeWrapper.cpp` - Read env var, set AppUserModelID
- `src/extractor/shortcut_helper.cpp` - NEW: COM helper for shortcuts
### Build
- `src/extractor/build.zig` - Link COM libraries
- `src/launcher/build.zig` - Console subsystem (fixes spinning cursor, no more deadlocks/threads during development))
### Bug Fixes
- `src/extractor/main.zig` - Remove spinner thread (Debug deadlock), add progress dots
- `src/launcher/main.zig` - Fix dev console visibility

## Testing
**Environment:** Windows 10 Pro
**Results:**
✅ **Pinning works correctly** - single unified taskbar icon, no duplicates, unique ID (from logs) = each Electrobun app has its own identity
✅ Debug mode no deadlock
✅ Dev console shows correctly in dev mode
✅ No breaking changes (Linux/macOS should be unaffected)

https://github.com/user-attachments/assets/333be9ec-566b-4143-9d3a-bdcaf62cae7c

https://github.com/user-attachments/assets/8a0561e9-81ca-4128-ad68-be1cac385f96

### DIFF
--- a/package/src/extractor/build.zig
+++ b/package/src/extractor/build.zig
@@ -18,5 +18,17 @@ pub fn build(b: *std.Build) void {
     // Use Console subsystem on all platforms so users can see extraction progress
     // The console window will automatically close when extraction completes
 
+    if (target.result.os.tag == .windows) {
+        exe.addCSourceFile(.{
+            .file = b.path("shortcut_helper.cpp"),
+            .flags = &.{"-std=c++17"},
+        });
+        exe.linkSystemLibrary("ole32");
+        exe.linkSystemLibrary("shell32");
+        exe.linkSystemLibrary("shlwapi");
+        exe.linkSystemLibrary("propsys");
+        exe.linkLibCpp();
+    }
+
     b.installArtifact(exe);
 }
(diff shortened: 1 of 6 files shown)


YoavCodes (changes requested):
I like the idea and appreciate the videos. I'd prefer it if this PR only had the Taskbar grouping changes.

When you get a chance could you resolve the merge conflicts and maybe break the other stuff out into another issue or PR.

YoavCodes:

PR #184, opened 2026-02-22
fix(types): extract ContextMenuItemConfig to remove role support from context menus

omarghandour:
Resolves a TODO inside `ContextMenu.ts`.

Context Menus should not share the exact same configuration type as Native Application Menus because application menus support the `role` property whereas context menus do not. This PR extracts a shared `BaseMenuItemConfig` and separates the `ApplicationMenuItemConfig` and `ContextMenuItemConfig` to restrict `role` usage to application menus only.

Tested locally via the `kitchen` sink application to ensure context menus still compile and render successfully.


YoavCodes:
> Resolves a TODO inside ContextMenu.ts.

I appreciate the PR, but I think the Todo was actually stale/incorrect.

updated it in https://github.com/blackboardsh/electrobun/commit/990c468ae5ed1a4dff0e8e74d0cd5ec772c1d275

PR #185, opened 2026-02-22
Add Flatpak support for Linux

candrewlee14:
Adds end-to-end Flatpak support for Linux builds.

## What's in here

**Native wrapper (`nativeWrapper.cpp`)**
- File choosers use `GtkFileChooserNative` so they route through `org.freedesktop.portal.FileChooser` in the sandbox
- `openFileManager()` uses `g_app_info_launch_default_for_uri()` (GIO/portal-aware) instead of `system("xdg-open ...")`
- `openURL()` drops the `system()` fallback — it doesn't work inside a sandbox anyway
- `showNotification()` calls `org.freedesktop.portal.Notification` over D-Bus instead of `system("notify-send ...")`
- Tray icon falls back to the theme icon name (`$FLATPAK_ID`) when running inside Flatpak, so the host compositor can find it
- XDG base dirs (`XDG_CACHE_HOME`, `XDG_DATA_HOME`, `XDG_RUNTIME_DIR`) take precedence over hardcoded `~/.cache` etc.

**CLI (`cli/index.ts`)**
- New `flatpak` config section under `build.linux`: `enabled`, `runtimeVersion`, `categories`, `extraFinishArgs`
- `createFlatpakBundle()` generates a Flatpak manifest, `.desktop` file, and AppStream metainfo, then runs `flatpak-builder` and `flatpak build-bundle` to produce a `.flatpak` artifact
- Uses `org.gnome.Platform` (includes `libwebkit2gtk-4.1`) with default runtime version `"48"`
- `collectExtraLibs()` uses `readelf -d` + `ldconfig -p` to walk the dep tree of `libNativeWrapper.so`, bundling only libs not provided by the runtime (in practice: the ayatana appindicator stack). Uses direct DT_NEEDED edges rather than `ldd`'s full transitive closure to avoid pulling in WebKit's ~100 deps and hitting ABI version conflicts with the runtime's own copies.
- Gracefully skips if `flatpak`/`flatpak-builder` aren't installed

**Kitchen config**
- `flatpak.enabled: true` added as a live example and to keep the feature tested

## Runtime requirements (build host)
- `flatpak` and `flatpak-builder`
- `readelf` (binutils)
- `org.gnome.Platform//48` and `org.gnome.Sdk//48` installed via Flathub

### DIFF
--- a/package/src/native/linux/nativeWrapper.cpp
+++ b/package/src/native/linux/nativeWrapper.cpp
@@ -4908,15 +4908,22 @@ void initializeGTK() {
     {
         std::unique_lock<std::mutex> lock(g_gtkInitMutex);
         if (!g_gtkInitialized) {
-            // Force X11 backend on Wayland systems
-            setenv("GDK_BACKEND", "x11", 1);
-            
+            // CEF requires X11 (no native Wayland support). When CEF is
+            // bundled, force GDK_BACKEND=x11 so everything runs under
+            // XWayland. When CEF is absent, let GTK choose the best backend
+            // (Wayland-native on a Wayland session, X11 otherwise).
+            if (isCEFAvailable()) {
+                setenv("GDK_BACKEND", "x11", 1);
+            }
+
             // Disable setlocale before gtk_init to prevent CEF conflicts
             gtk_disable_setlocale();
             gtk_init(nullptr, nullptr);
-            
-            // Install X11 error handler for debugging
-            XSetErrorHandler(x11_error_handler);
+
+            // Install X11 error handler — only valid for X11 displays
+            if (GDK_IS_X11_DISPLAY(gdk_display_get_default())) {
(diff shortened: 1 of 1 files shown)


candrewlee14:
@YoavCodes I did some manual testing here on getting the flatpak running, and all the browser tests passed on it.

YoavCodes:
Thanks, might consider Flatpak in the future, but for now it’s not aligned with Electrobun’s direction:

  - Requires users to install/configure Flatpak and runtimes/remotes.
  - Adds packaging/runtime overhead and often larger distribution footprint than our Ubuntu-targeted flow.
  - Conflicts with our batteries-included updater model (which generates very small differential patches as small as 4KB).
  - Introduces hardcoded dependency logic in the CLI that creates ongoing maintenance risk.

  Electrobun currently supports recent Ubuntu versions only, so this complexity isn’t justified right now.

Issue #191, opened 2026-02-23
setApplicationMenu fails silently — dispatch_async reads freed buffer (dist/ copy has no fix)


stefandevo:
## Summary

`ApplicationMenu.setApplicationMenu()` silently fails — the menu bar shows only the app name with no subitems or additional menus (Edit, Window, etc.). The native Obj-C layer logs:

```
Setting application menu from JSON in objc
Failed to parse JSON: Error Domain=NSCocoaErrorDomain Code=3840
  "Unable to parse empty data." UserInfo={NSDebugDescription=Unable to parse empty data.}
```

This is related to #136 and #160 but with additional findings.

## Environment

- electrobun 1.13.1
- macOS (Apple Silicon)
- Bun runtime
- App uses top-level `await` (async module)

## Root Cause

There are **two issues**:

### 1. `dist/` vs `dist-macos-arm64/` — the runtime file has no fix

The `package.json` exports map points to `./dist/api/bun/index.ts`:

```json
"exports": {
    ".": "./dist/api/bun/index.ts",
    "./bun": "./dist/api/bun/index.ts",
}
```

The `dist/api/bun/proc/native.ts` file still uses the original unfixed code:

```typescript
setApplicationMenu: (params: { menuConfig: string }): void => {
    const { menuConfig } = params;
    native.symbols.setApplicationMenu(
        toCString(menuConfig),   // ← ephemeral buffer, GC'd immediately
        applicationMenuHandler,
    );
},
```

Meanwhile, `dist-macos-arm64/api/bun/proc/native.ts` has a buffer-pinning fix with `globalThis.__electrobun_menuConfigBuf` — but **this file is never loaded at runtime** because the exports map doesn't point to it.

### 2. Buffer pinning alone is insufficient

Even with the `dist-macos-arm64` fix applied manually to `dist/`, the menu still fails. The fix uses:

```typescript
const buf = Buffer.from(menuConfig + "\0", "utf8");
(globalThis as any).__electrobun_menuConfigBuf = buf;
native.symbols.setApplicationMenu(ptr(buf), applicationMenuHandler);
```

This keeps the `Buffer` object alive, but `Buffer.from(string)` in Bun may use a **pooled slab allocator** for small strings. The `ptr(buf)` pointer addresses a region within a shared slab — and subsequent `Buffer.from()` calls (from other parts of the app executing before `dispatch_async` runs) can overwrite that slab region, even though the original Buffer object is still referenced.

[...]

Issue #136, opened 2026-02-05
ApplicationMenu.setApplicationMenu fails on macOS if main script uses async/await


p2k:
Steps to reproduce (macOS Tahoe 26.2 on an M1 Pro, could reproduce with electrobun 0.8.0, 1.11.3-beta.1, 1.11.4-beta.0):

```
bunx electrobun init
# follow the instructions, choose the hello-world template
cd my-hello-world-app
```

Then edit src/bun/index.ts so it looks like this:
```
import { BrowserWindow, ApplicationMenu } from "electrobun/bun";

// Simple application menu
ApplicationMenu.setApplicationMenu([
  {
    submenu: [
      { role: "quit", accelerator: "q" },
    ],
  },
]);

// Example await call
await new Promise(resolve => setTimeout(resolve, 1000));

// Create the main application window
const mainWindow = new BrowserWindow({
  title: "Hello Electrobun!",
  url: "views://mainview/index.html",
  frame: {
    width: 800,
    height: 800,
    x: 200,
    y: 200,
  },
});

console.log("Hello Electrobun app started!");
```

Then run:
```
bun install
bun start
```

The logs will say something like:
```
Setting application menu from JSON in objc
Failed to parse JSON: Error Domain=NSCocoaErrorDomain Code=3840 "Unable to parse empty data." UserInfo={NSDebugDescription=Unable to parse empty data.}
```

I found this issue while playing with a modified version of the svelte template (I changed it so it uses Vue.js instead). In that template, it uses an async function to check if we run the dev version.

Just to be clear: if this issue can't be fixed, I strongly recommend updating the documentation so it clarifies that the main script cannot have any async/await if macOS is a target. Also `ApplicationMenu.setApplicationMenu` call must be made _before_ any BrowserWindow is created or it will likewise fail/be ignored.

pedropalhari:
For reference while the PR is still open, if you call `await` between `ApplicationMenu.setApplicationMenu(...)` and `new BrowserWindow` it fails. But if you move the `await` call before the `ApplicationMenu.setApplicationMenu(...)` it works properly.

On the default React + Tailwind + Vite template, you just can work around this by calling `.setApplicationMenu` right after the `await` to get the main view URL.

YoavCodes:
Fixed in the next stable release.

PR #192, opened 2026-02-23
fix: copy JSON string before dispatch_async in menu functions

### LINKED ISSUE #160: setApplicationMenu: dangling pointer in dispatch_async causes silent JSON parse failure (menu never created)
## Bug

`ApplicationMenu.setApplicationMenu()` silently fails to create any menu. The native menu bar only shows the app name with no custom menus, and keyboard accelerators never fire.

## Root Cause

`toCString()` in `dist/api/bun/proc/native.ts` creates a `Buffer` and returns a raw pointer via `ptr(buf)`. The `Buffer` itself is not stored anywhere — it's a local temporary eligible for GC immediately after the FFI call returns.

The native `setApplicationMenu` in `nativeWrapper.mm` does this:

```objc
extern "C" void setApplicationMenu(const char *jsonString, ...) {
    NSLog(@"Setting application menu from JSON in objc");
    dispatch_async(dispatch_get_main_queue(), ^{
        NSData *jsonData = [NSData dataWithBytes:jsonString length:strlen(jsonString)];
        // ^ jsonString pointer is already dangling here
    });
}
```

The `dispatch_async` block captures `jsonString` **by pointer**, not by value. By the time the block executes on the main thread, Bun's GC has already freed the `Buffer` backing the pointer. `strlen()` on freed memory returns 0, `NSData` gets 0 bytes, and `NSJSONSerialization` fails with:

```
Failed to parse JSON: Error Domain=NSCocoaErrorDomain Code=3840 "Unable to parse empty data."
```

The menu is silently never created. No crash, no other error.

## Diagnosis Trail

The only visible symptoms are:
1. Menu bar shows only the app name (no custom menus)
2. This log line appears ~20ms after "Setting application menu from JSON in objc":
   `Failed to parse JSON: Error Domain=NSCocoaErrorDomain Code=3840 "Unable to parse empty data."`
3. The `application-menu-clicked` event never fires for any action or accelerator

## Fix

Keep a reference to the `Buffer` alive until the main thread has consumed it. Minimal JS-side fix in `dist/api/bun/proc/native.ts`:

```diff
 setApplicationMenu: (params: { menuConfig: string }): void => {
     const { menuConfig } = params;
+    // Prevent GC from freeing the buffer before dispatch_async runs on main thread
+    const buf = Buffer.from(menuConfig + "\0", "utf8");
+    (globalThis as any).__electrobun_menuConfigBuf = buf;
     native.symbols.setApplicationMenu(
-        toCString(menuConfig),
+        ptr(buf),
         applicationMenuHandler,
     );
 },
```

The proper native-side fix would be to copy the JSON string into an `NSString` **before** `dispatch_async`:

```objc
extern "C" void setApplicationMenu(const char *jsonString, ZigStatusItemHandler zigTrayItemHandler) {
    NSLog(@"Setting application menu from JSON in objc");
    NSString *jsonCopy = [NSString stringWithUTF8String:jsonString]; // copy while pointer is valid
    dispatch_async(dispatch_get_main_queue(), ^{
        NSData *jsonData = [jsonCopy dataUsingEncoding:NSUTF8StringEncoding];
        // ... rest unchanged
    });
}
```

The same pattern likely affects `showContextMenu` and `setTrayMenuFromJSON` but may be less reproducible due to timing differences.

## Environment

- electrobun: 1.12.3
- macOS: 15.5 arm64
- Bun: (via electrobun bundled runtime)

stefandevo:
## Summary

- Copy `const char *jsonString` into an `NSString` (ARC-managed) **before** `dispatch_async` in all three affected functions: `setApplicationMenu`, `showContextMenu`, and `setTrayMenuFromJSON`
- The dispatch block now captures the `NSString` copy, which is retained by ARC regardless of Bun's GC timing

## Problem

The native Obj-C functions receive a `const char *` from Bun's FFI and pass it into a `dispatch_async(dispatch_get_main_queue(), ...)` block. The block captures the raw pointer by value. By the time the main thread dequeues and executes the block, Bun's garbage collector has freed or overwritten the memory at that address, resulting in:

```
Failed to parse JSON: Error Domain=NSCocoaErrorDomain Code=3840
  "Unable to parse empty data."
```

The menu bar shows only the app name with no subitems or additional menus.

## Root cause

`toCString()` in the JS layer creates a `Buffer.from(string)` and returns `ptr(buf)`. The `Buffer` is a local variable that becomes GC-eligible immediately after the FFI call returns. For synchronous native calls this is fine (the native code reads the string before returning), but `dispatch_async` defers the read to a later main-thread tick — by which time the buffer is gone.

A JS-side workaround (pinning the buffer to `globalThis`) was attempted in `dist-macos-arm64` but:
1. The `dist/` copy (which is what `package.json` exports actually resolve to) was never patched
2. Even with pinning, `Buffer.from()` may use Bun's pooled slab allocator, and subsequent buffer allocations can overwrite the slab region

## Fix

The proper fix is on the native side — one line per function:

```objc
NSString *jsonCopy = [NSString stringWithUTF8String:jsonString];
```

Called **before** `dispatch_async`, so the block captures an ARC-retained `NSString` instead of a raw `const char *`. The `NSData` is then created from the copy inside the block:

```objc
NSData *jsonData = [jsonCopy dataUsingEncoding:NSUTF8StringEncoding];
```

## Affected functions

| Function | File | Line |
|----------|------|------|
| `setApplicationMenu` | `nativeWrapper.mm` | ~6480 |
| `showContextMenu` | `nativeWrapper.mm` | ~6499 |
| `setTrayMenuFromJSON` | `nativeWrapper.mm` | ~6449 |

## Test plan

- [x] Verified fix resolves the issue in a real Electrobun app (DevFlow) using `setApplicationMenu` with top-level `await` in the entry point
- [x] Build native dylib and verify menus appear correctly
- [x] Verify context menus still work
- [x] Verify tray menus still work

Fixes #160, #136, #191

### DIFF
--- a/package/src/native/macos/nativeWrapper.mm
+++ b/package/src/native/macos/nativeWrapper.mm
@@ -6447,10 +6447,12 @@ extern "C" void setTrayImage(NSStatusItem *statusItem, const char *image) {
 
 
 extern "C" void setTrayMenuFromJSON(NSStatusItem *statusItem, const char *jsonString) {
+    // Copy the JSON string before dispatch_async (same GC race as setApplicationMenu)
+    NSString *jsonCopy = [NSString stringWithUTF8String:jsonString];
     dispatch_async(dispatch_get_main_queue(), ^{
         if (statusItem) {
             StatusItemTarget *target = objc_getAssociatedObject(statusItem.button, "statusItemTarget");
-            NSData *jsonData = [NSData dataWithBytes:jsonString length:strlen(jsonString)];
+            NSData *jsonData = [jsonCopy dataUsingEncoding:NSUTF8StringEncoding];
             NSError *error;
             NSArray *menuArray = [NSJSONSerialization JSONObjectWithData:jsonData options:0 error:&error];
             if (error) {
@@ -6479,8 +6481,12 @@ extern "C" void removeTray(NSStatusItem *statusItem) {
 
 extern "C" void setApplicationMenu(const char *jsonString, ZigStatusItemHandler zigTrayItemHandler) {
     NSLog(@"Setting application menu from JSON in objc");
+    // Copy the JSON string before dispatch_async. The original pointer comes
+    // from Bun's FFI and may be freed/overwritten by JS GC before the main
+    // thread dequeues this block.
+    NSString *jsonCopy = [NSString stringWithUTF8String:jsonString];
     dispatch_async(dispatch_get_main_queue(), ^{
-        NSData *jsonData = [NSData dataWithBytes:jsonString length:strlen(jsonString)];
+        NSData *jsonData = [jsonCopy dataUsingEncoding:NSUTF8StringEncoding];
         NSError *error;
         NSArray *menuArray = [NSJSONSerialization JSONObjectWithData:jsonData options:0 error:&error];
         if (error) {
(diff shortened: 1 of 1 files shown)


stefandevo:
Any chance this gets merged?

YoavCodes:
Appreciate the PR, but this was already fixed by the time i got here. nice work though! ✨

PR #271, opened 2026-03-08
fix: enable persistent cookie storage on Linux (WebKitGTK)

### LINKED ISSUE #186: Session/cookies not working on linux
According to the [docs](https://blackboard.sh/electrobun/docs/apis/utils/), the following should work:

```ts
const partition = 'persist:test'
const session = Session.fromPartition(partition)

  title: `Test`,
  url: 'https://setcookie.net',
  partition: partition,
  frame: { width: 800, height: 600, x: 100, y: 100 }
})

win.on('close', () => {
  console.log(session.cookies.get())
})
```

However this always returns/prints `[]`, regardless of cookies having been set.

alexandreolives:
## Summary

- Call `webkit_cookie_manager_set_persistent_storage()` after creating `WebKitWebsiteDataManager` for persistent partitions in `getContextForPartition()` and `getDataManagerForPartition()`
- Cookies are stored as SQLite in the partition's data directory (`cookies.sqlite`)
- Only affects persistent partitions (`persist:` prefix) — ephemeral partitions are unchanged

## Problem

On Linux, cookies are kept in memory only and lost on restart. `Session.fromPartition(partition).cookies.get()` always returns `[]`.

The root cause is that `webkit_cookie_manager_set_persistent_storage()` is never called in the native GTK wrapper. Without this call, WebKitGTK defaults to in-memory cookie storage.

## Fix

After creating a `WebKitWebsiteDataManager` for a persistent partition, get its cookie manager and configure SQLite-backed persistent storage:

```cpp
WebKitCookieManager* cookieManager = webkit_website_data_manager_get_cookie_manager(dataManager);
if (cookieManager) {
    std::string cookiePath = dataPath + "/cookies.sqlite";
    webkit_cookie_manager_set_persistent_storage(
        cookieManager, cookiePath.c_str(),
        WEBKIT_COOKIE_PERSISTENT_STORAGE_SQLITE
    );
}
```

Fixes #186

### DIFF
--- a/package/src/native/linux/nativeWrapper.cpp
+++ b/package/src/native/linux/nativeWrapper.cpp
@@ -2648,7 +2648,7 @@ public:
                 gtk_widget_set_size_request(wrapper, frame.width, frame.height);
                 gtk_widget_set_margin_start(wrapper, clampedX);
                 gtk_widget_set_margin_top(wrapper, clampedY);
-                
+
                 // Position webview within wrapper with offset to handle negative positions
                 // Note: /2 division appears necessary for GTK coordinate system
                 gtk_fixed_move(GTK_FIXED(wrapper), webview, offsetX / 2, offsetY / 2);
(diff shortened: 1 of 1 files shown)


alexandreolives:
Apologies for the confusion — I reorganized branches to have one clean PR per fix instead of everything on a single branch. This PR's branch was deleted in the process.

The same fix is now in #275 with a dedicated branch. Sorry for the noise!

Issue #272, opened 2026-03-08
Linux: secondary BrowserView added as 1x1 GtkFixed overlay (no scroll/resize)


alexandreolives:
## Description

On Linux (GTK/WebKitGTK renderer), when creating a second `BrowserView` attached to an existing `BrowserWindow`, the webview is added as an overlay inside a `GtkFixed` container with `gtk_widget_set_size_request(wrapper, 1, 1)`.

This makes the secondary webview effectively 1x1 pixel — scroll doesn't work and the content is clipped.

## Root cause

In `nativeWrapper.cpp`, the first webview is added directly to the container (`gtk_container_add`), but subsequent webviews go through a different code path:

```cpp
// ~L4540-4568
GtkWidget* wrapper = gtk_fixed_new();
gtk_widget_set_size_request(wrapper, 1, 1); // <-- problem: 1x1 px
gtk_widget_set_events(wrapper, 0);
gtk_widget_set_can_focus(wrapper, FALSE);
gtk_fixed_put(GTK_FIXED(wrapper), view->widget, 0, 0);
gtk_overlay_add_overlay(GTK_OVERLAY(overlay), wrapper);
```

The `resize()` method does try to update size, but with the issues from #188, the wrapper can only grow, never shrink.

## Impact

This prevents use cases like:
- Replacing the default webview with a partitioned one (for cookie isolation)
- Multi-tab or multi-panel layouts on Linux
- Any scenario requiring more than one BrowserView per window

## Steps to reproduce

```typescript
const win = new BrowserWindow({ url: "about:blank", frame: { width: 1280, height: 800 } });

const newView = new BrowserView({
  url: "https://example.com",
  windowId: win.id,
  frame: { x: 0, y: 0, width: 1280, height: 800 },
});

// The default webview works fine (scroll, resize)
// The new BrowserView is stuck at 1x1 pixel
```

## Suggested fix

Use `gtk_widget_set_size_request(wrapper, frame.width, frame.height)` instead of `(wrapper, 1, 1)`, and ensure the wrapper participates in the window's resize flow (similar to the first webview).

## Environment

- OS: Linux Mint 22 (X11)
- GTK/WebKitGTK (native renderer)
- Electrobun 1.15.1

Related: #188 (resize ratchet), #253 (multi-tab issues)

alexandreolives:
**Fix PR:** #274

[...]

PR #274, opened 2026-03-08
fix: add full-size BrowserView as proper overlay

alexandreolives:
## Summary
- When adding a subsequent BrowserView with `fullSize=true`, it was wrapped in a 1x1 `GtkFixed` container meant for OOPIFs
- This caused the webview to not receive scroll events and not resize properly
- Full-size webviews are now added directly as `GtkOverlay` children with `expand=TRUE`, matching the behavior of the first webview
- The original `GtkFixed` wrapper is preserved for non-fullSize OOPIF views

## Test plan
- [x] Build and run on Linux (WebKitGTK)
- [x] Verify second BrowserView (e.g. partitioned webview) scrolls and resizes correctly
- [x] Verify OOPIF behavior is unchanged

Related: #272


YoavCodes (approved):
Thanks!

Issue #261, opened 2026-03-06
expose env in electrobun.config.ts


theoephraim:
Im trying to toggle the application name and icon based on what env im in - this is useful for example so I can easily tell if I'm running the dev build or not.

The docs say there are env vars set in hook scripts, but I guess not within `electrobun.config.ts`?

is there another way to accomplish this?

PR #279, opened 2026-03-08
Allow codesigning in dev builds with ad-hoc fallback

theoephraim:
## Summary
- Remove the `buildEnvironment !== "dev"` gate from `shouldCodesign`, so dev builds can be codesigned when `codesign: true` is set in config
- Fall back to ad-hoc signing (`--sign "-"`) when `ELECTROBUN_DEVELOPER_ID` env var is not set, instead of exiting with an error. Ad-hoc signing skips `--timestamp` and `--options runtime` (which require a real Developer ID cert)
- Fix missing `dirname` import in the WebGPU FFI generator script (`gen-webgpu-ffi.mjs`)

## Motivation
Ad-hoc codesigning in dev mode is needed to test macOS permission prompts (e.g., microphone access) which require the app to be signed with appropriate entitlements. Previously, codesigning was completely blocked in dev builds.

## Notes
Also somewhat related to https://github.com/blackboardsh/electrobun/issues/261 - as you may want to manually control config based on the current env.

## Test plan
- [ ] Run `electrobun dev` with `codesign: true` and **without** `ELECTROBUN_DEVELOPER_ID` set — should ad-hoc sign successfully
- [ ] Run `electrobun dev` with `codesign: true` and **with** `ELECTROBUN_DEVELOPER_ID` set — should sign with real identity as before
- [ ] Run `electrobun dev` with `codesign: false` — should skip signing as before
- [ ] Verify `codesign --verify --deep --strict` passes on the built .app
- [ ] Verify macOS permission prompts (e.g., microphone) work with ad-hoc signed dev build


YoavCodes:

PR #282, opened 2026-03-09
fix: add missing @types/three dependency

### LINKED ISSUE #280: v1.15.1: Missing @types/three causes type errors for consumers
## Problem

Since v1.15.1, `electrobun/dist/api/bun/index.ts` imports `three` (added in a6fecd6 ), but `@types/three` is not included in `dependencies`. This causes `TS7016` errors when consumers run type checking.

As noted in #117, `skipLibCheck: true` does not help because electrobun ships `.ts` source files, not `.d.ts`.

## Steps to Reproduce

- Install `electrobun@1.15.1`
- Run `tsgo --noEmit` (or `tsc --noEmit`) with `strict: true`

## Error

```
node_modules/electrobun/dist/api/bun/index.ts(32,24): error TS7016: Could not find a declaration file for module 'three'.
```

## Suggested Fix

Add `@types/three` to `dependencies` in `package/package.json`, alongside `three` — the same way `@types/bun` is already included in `dependencies`.

`@babylonjs/core` (also added in the same commit) ships its own type declarations and does not have this issue.

bigmacfive:
## Summary
- add `@types/three` to package dependencies so consumers typechecking Electrobun's shipped TypeScript sources do not hit `TS7016`
- update the Bun lockfile

## Testing
- `cd package && bun install`
- `git diff --check`
- `bun run typecheck` is currently blocked by pre-existing missing generated files (`src/bun/proc/native.ts`, `src/cli/index.ts`)

Closes #280

### DIFF
--- a/package/bun.lock
+++ b/package/bun.lock
@@ -7,6 +7,7 @@
       "dependencies": {
         "@babylonjs/core": "^7.45.0",
         "@types/bun": "^1.3.8",
+        "@types/three": "^0.165.0",
         "png-to-ico": "^2.1.8",
         "proxy-agent": "^6.5.0",
         "rcedit": "^4.0.1",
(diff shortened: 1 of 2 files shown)


aleclarson:
Electrobun should compile its `.ts` to `.js`/`.d.ts` artifacts and publish those, instead of burdening the downstream type-checker with unnecessary work.

YoavCodes:

PR #284, opened 2026-03-09
fix: use direct child paths for CEF partitions

### LINKED ISSUE #278: CEF persistent partitions fail on macOS because profile path is nested under Partitions/
## Summary

Persistent CEF partitions on macOS fail to initialize because Electrobun creates the partition profile under a nested path like:

`<root_cache_path>/Partitions/<partition-name>`

For example:

`~/Library/Application Support/<identifier>/<channel>/CEF/Partitions/default`

CEF then fails with:

```text
ERROR:cef/libcef/browser/chrome/chrome_browser_context.cc:116] Cannot create profile at path .../CEF/Partitions/default
```

## Why this fails

Electrobun sets:

- `root_cache_path = .../<identifier>/<channel>/CEF`
- `cache_path = .../<identifier>/<channel>/CEF/Partitions/<partition-name>`

For Chrome runtime CEF, the profile path used for a persistent `CefRequestContext` must be the root cache dir itself or an immediate child of it. The nested `Partitions/<name>` layout is rejected during profile creation.

This means the problem is not:

- file permissions
- app identifier naming
- a specific partition name
- dev vs stable channel

It affects any `persist:*` partition, including `persist:default`.

## Reproduction

On macOS, create a CEF webview with a persistent partition such as:

- `persist:default`
- `persist:test1`

Electrobun logs:

```text
DEBUG CEF: CreateRequestContextForPartition called for webview 1, partition: persist:default
```

Then CEF logs:

```text
ERROR:cef/libcef/browser/chrome/chrome_browser_context.cc:116] Cannot create profile at path /Users/.../Library/Application Support/<identifier>/<channel>/CEF/Partitions/default
```

## Root cause in Electrobun

`buildPartitionPath(...)` currently generates:

`.../<identifier>/<channel>/<renderer>/Partitions/<partition-name>`

That path is used for CEF partition `cache_path` in:

- `package/src/native/macos/nativeWrapper.mm`
- `package/src/native/linux/nativeWrapper.cpp`
- `package/src/native/win/nativeWrapper.cpp`
- helper in `package/src/native/shared/app_paths.h`

## Why the fix works

Changing the CEF partition path from:

`.../CEF/Partitions/<partition-name>`

to:

`.../CEF/<partition-name>`

fixes the issue, because the partition profile becomes a direct child of `root_cache_path`, which matches what CEF expects for persistent profile directories.

After that change, persistent partitions such as `persist:default` and `persist:test1` initialize correctly.

## Suggested fix

For CEF only, build partition paths as:

`<root_cache_path>/<partition-name>`

instead of:

`<root_cache_path>/Partitions/<partition-name>`

Other renderers can keep their existing path layout if they do not have the same restriction.

bigmacfive:
## Summary
- add a CEF-specific partition path helper that places persistent profile directories directly under the renderer root cache path
- switch the macOS, Linux, and Windows CEF request-context callsites to use that helper

## Testing
- `git diff --check`
- did not run native platform repros in this environment

Closes #278

### DIFF
--- a/package/src/native/linux/nativeWrapper.cpp
+++ b/package/src/native/linux/nativeWrapper.cpp
@@ -3587,7 +3587,7 @@ CefRefPtr<CefRequestContext> CreateRequestContextForPartition(const char* partit
             // Build cache path with identifier/channel structure (consistent with CLI and updater)
             char* home = getenv("HOME");
             std::string basePath = home ? std::string(home) + "/.cache" : "/tmp";
-            std::string cachePath = buildPartitionPath(basePath, g_electrobunIdentifier, g_electrobunChannel, "CEF", partitionName);
+            std::string cachePath = buildCEFPartitionPath(basePath, g_electrobunIdentifier, g_electrobunChannel, "CEF", partitionName);
 
             // Create directory
             g_mkdir_with_parents(cachePath.c_str(), 0755);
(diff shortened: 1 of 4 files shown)


LeoNilsson:
Any idea of when this will be reviewed / released?

YoavCodes (approved):
Thanks!

PR #301, opened 2026-03-11
Add `dev.reloadWithoutActivating` option for focus-free live reload

### LINKED ISSUE #288: option to disable automatic stealing focus on dev launch
Doing dev tasks with live reload on means the app can be relaunching itself many times. Each time it is stealing focus on my machine which can be quite disruptive. While it is a sane default, an option to disable this could be good. Or some additional smart logic - something like "if app hasnt been relaunched in last 5 minutes, steal focus, otherwise dont".

theoephraim:
Closes #288

The problem is that while working on my app, its constantly reloading and stealing focus.
This new option lets us launch normally in dev mode, and then re-launch without stealing focus on reloads.

## Summary
- Adds a native `showWindowWithoutActivating` function (macOS) that uses `orderFrontRegardless` to show windows without stealing keyboard focus
- Exposes it through FFI, the request layer, and as `BrowserWindow.showWithoutActivating()` / `GpuWindow.showWithoutActivating()`
- Adds a `dev.reloadWithoutActivating` config option that automatically uses the non-activating path during live-reload rebuilds (`electrobun dev --watch`), while the initial launch still activates normally

## Usage
```ts
// electrobun.config.ts
export default {
  dev: {
    reloadWithoutActivating: true,
  },
} satisfies ElectrobunConfig;
```

## Test plan
- [ ] Run `electrobun dev --watch` with `reloadWithoutActivating: true` — verify initial launch steals focus
- [ ] Make a file change to trigger live-reload — verify the app window reappears without stealing focus from the editor
- [ ] Run `electrobun dev --watch` with `reloadWithoutActivating: false` (or omitted) — verify reload still steals focus as before
- [ ] Call `win.showWithoutActivating()` manually from app code — verify it works independently of the config option

### DIFF
--- a/package/src/bun/ElectrobunConfig.ts
+++ b/package/src/bun/ElectrobunConfig.ts
@@ -328,11 +328,12 @@ export interface ElectrobunConfig {
 	 */
 	dev?: {
 		/**
-		 * Show windows without stealing focus from the currently active app.
-		 * Useful during development so live-reload doesn't yank you out of your editor.
+		 * When live-reloading (`electrobun dev --watch`), show windows without
+		 * stealing focus from the currently active app. The initial launch still
+		 * activates normally.
 		 * @default false
 		 */
-		launchWithoutActivating?: boolean;
+		reloadWithoutActivating?: boolean;
 	};
 
 	/**
--- a/package/src/bun/ElectrobunConfig.ts
+++ b/package/src/bun/ElectrobunConfig.ts
@@ -323,6 +323,19 @@ export interface ElectrobunConfig {
 	};
 
 	/**
+	 * Dev mode configuration.
+	 * Controls behaviour when running with `electrobun dev`.
+	 */
+	dev?: {
+		/**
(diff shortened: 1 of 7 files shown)


YoavCodes:
Thanks for the PR. I implemented this directly on `main` with a different API shape and native behavior, so I’m closing this one.

with the new api if you want to disabling auto-activating windows you can set activate:false in the window options or use showInactive() instead of show() for a particular window while working with that/those windows.

doing it app-wide from electrobun.config.ts doesn't make sense to me.

theoephraim:
No worries. I think my main issue was that I was happy with having it automatically activate the window when I first boot in dev mode, but skip it when on each live reload, as that became very disruptive. If I can programatically detect and control that, then I'd be happy. Since it was purely a dev concern, the config file kind of made sense.

Anyway I've been working on other stuff but hopefully coming back to this project soon so will report back if it still feels weird.

Thanks!

PR #311, opened 2026-03-16
fix: explicitly resolve rcedit from dependency tree for Windows packaging

### LINKED ISSUE #298: Windows packaging resolves rcedit from a baked CI path in electrobun@1.15.1
# ElectroBun issue draft

## Title

Windows packaging resolves `rcedit` from a baked CI path in `electrobun@1.15.1`

## Summary

`electrobun@1.15.1` packages successfully on Windows, but icon embedding tries to execute `rcedit` from a baked CI path like:

`D:\a\electrobun\electrobun\package\node_modules\rcedit\bin\rcedit-x64.exe`

That path does not exist in normal local installs, so Windows packaging falls back to warnings and drops the configured app icon.

## Reproduction

Environment:

- Windows
- `electrobun@1.15.1`
- local app project using `build.win.icon`

Steps:

1. Install dependencies in a normal app project that depends on ElectroBun.
2. Configure `build.win.icon` in `electrobun.config.ts`.
3. Run `electrobun build --env=stable`.

## Actual behavior

ElectroBun logs warnings similar to:

```text
Warning: Failed to embed icon into launcher.exe: Error executing command (D:\a\electrobun\electrobun\package\node_modules\rcedit\bin\rcedit-x64.exe ...)
spawn D:\a\electrobun\electrobun\package\node_modules\rcedit\bin\rcedit-x64.exe ENOENT
```

The package may still build, but the icon embedding step does not use the configured icon.

## Expected behavior

ElectroBun should resolve `rcedit` from the locally installed dependency tree and embed the configured icon successfully on Windows.

## Root cause

The Windows icon embedding path uses `rcedit` in a way that ends up tied to a baked build-time/snapshot path instead of the local package installation path.

## Proposed fix

In `src/cli/index.ts`, replace the current `await import("rcedit")` usage in the Windows icon embedding paths with explicit local resolution:

1. Resolve `rcedit/package.json` from the project dependency tree.
2. Derive the installed `rcedit` package directory from that path.
3. Prefer `bin/rcedit-x64.exe`, then fallback to `bin/rcedit.exe`.
4. Execute that resolved binary directly for `--set-icon`.
5. Keep the existing PNG-to-ICO conversion and warning-only failure handling.

## Attached repo artifacts

- Source diff: `docs/electrobun-rcedit-upstream.diff`
- Notes: `docs/electrobun-rcedit-upstream.md`

Sigmabrogz:
## Summary
Resolves `rcedit` explicitly to avoid falling back to the baked CI path during Windows packaging.

## Problem
Closes #298
When running `electrobun build` for Windows, it tries to execute `rcedit` from a baked CI path like `D:\a\electrobun\electrobun\package\node_modules\rcedit\bin\rcedit-x64.exe`, which fails on local environments.

## Solution
Replaced `await import("rcedit")` in `src/cli/index.ts` with explicit local dependency resolution:
- Resolves `rcedit/package.json` using `require.resolve`.
- Selects `rcedit-x64.exe` if available, falling back to `rcedit.exe`.
- Executes it via `execFileSync`.

## Testing
- Code modifications mirror the proposed fix in the issue closely.

### DIFF
--- a/package/src/cli/index.ts
+++ b/package/src/cli/index.ts
@@ -2163,10 +2163,12 @@ Categories=Utility;Application;
 					}
 
 					// Use rcedit to embed the icon into launcher.exe
-					const rcedit = (await import("rcedit")).default;
-					await rcedit(bunCliLauncherDestination, {
-						icon: iconPath,
-					});
+										const { execFileSync } = await import("child_process");
+					const rceditPkgPath = require.resolve("rcedit/package.json");
+					const rceditDir = dirname(rceditPkgPath);
+					const rceditX64 = join(rceditDir, "bin", "rcedit-x64.exe");
+					const rceditExe = existsSync(rceditX64) ? rceditX64 : join(rceditDir, "bin", "rcedit.exe");
+					execFileSync(rceditExe, [bunCliLauncherDestination, "--set-icon", iconPath]);
 					console.log(`Successfully embedded icon into launcher.exe`);
 
 					// Clean up temp ICO file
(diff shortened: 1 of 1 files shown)


YoavCodes (approved):
Thanks!

PR #319, opened 2026-03-17
Windows: preserve native frame behavior for hidden titlebars and add Windows-specific custom titlebar interactions

### LINKED ISSUE #320: Windows not preserving native frame behavior for hidden titlebars
On Windows, using a custom titlebar currently does not behave as expected when switching away from the default native titlebar.

With the `default` titlebar, the window behaves correctly:
- Native borders and rounded corners are present
- The window can be resized from all expected edges
- Native minimize, maximize, and close animations are preserved

With `hiddenInset` on Windows:
- The titlebar is mostly removed, but a small part remains visible at the top
- Native borders remain and resizing still works
- Native close, minimize, and maximize animations are lost

With `hidden` on Windows:
- The titlebar is fully removed
- Native borders and rounded corners are lost
- Top-edge resize behavior is broken or missing
- Native close, minimize, and maximize animations are lost
- Window snapping does not work

This makes it difficult to build a Windows app with a fully custom title bar while still preserving the expected native window frame behavior.

Here is an image of how the  "Custom Titlebar Demo" looks on my Windows 11 machine, where it is not possible to resize the window. 

Image

I think a part of the problem is that `hiddenInset` appears to be treated as a meaningful option on Windows, even though to my understanding it is a macOS-style concept. On Windows, I think it would be cleaner if title bar visibility and border/frame behavior were controlled separately.

Therefore, I would like to suggest: 

- Remove `hiddenInset` as option on Windows, or just treat as 'hidden'
- Allow `hidden` to remove only the native titlebar while preserving the standard Windows frame behavior by default
- Use the separate borderless flag for explicitly removing borders and frame styling when that is actually desired

I made some changes using Codex and Claude, and tested a working implementation of this approach on Windows 11 using the Kitchen examples "Custom Titlebar Demo" and "Transparent/Borderless window for floating UI testing". Please see PR: #319

Andreas-Froyland:
This PR improves custom titlebar behavior on Windows so hidden titlebars behave much closer to native Windows windows, and is based on the feedback I sent in the [discord chat ](https://discordapp.com/channels/1190061229914464397/1468277801374453974/1478712384162893875), and would resolve #320 

The main change is separating titlebar visibility from frame removal. On Windows, `hiddenInset` is not really a native framing concept, so it should not drive window-frame behavior. Instead, `hidden` should remove the visible titlebar while preserving the native frame by default, and `borderless` should become the explicit opt-in for removing borders and frame styling when users want to have fully transparent overlays and similar.

I have also added some hidden-titlebar interaction behavior on Windows to better match the patterns used by standard Windows windows, inspired from how it is handled in Electron and Tauri.

What changes on Windows:
- Preserves native borders and rounded corners when using a hidden titlebar
- Restores top-edge resize handling for hidden/custom titlebar windows
- Preserves expected native minimize, maximize, and close animations
- Keeps border removal behind an explicit `borderless` path instead of making it an implicit side effect of hidden titlebars
- Supports double-click on the draggable titlebar region to maximize / restore to match Windows native windows behaviour
- Supports dragging from a maximized titlebar by restoring the window and continuing the drag. Also to match native windows window behavior. 

As noted in the GitHub issue as well, I sent Codex and Claude to fix it up based on how Tauri and Electron codebases does window creation on Windows for hidden titlebars, as I am not that familiar with the Windows API for window creation:) Feel free to close it or request changes if needed. 

We might want to rethink the requirement to set borderless true as well for fully custom transparent views, as it could possibly break existing windows. Maybe default it to true, and require to manually set it to false when wanting frames on a window?

Tested on Windows 11 using Kitchen:
- Custom Titlebar Demo
- Transparent/Borderless window used for floating UI testing

On my machine, both examples now behave as expected.

See images below for before and after: 

Before when using 'hidden': 

image

After when using 'hidden' and 'borderless: true' for transparent windows:

image


YoavCodes:

PR #321, opened 2026-03-17
fix: export the RESOURCES_FOLDER

bitporo:
When I look at the official website, found that there is this field, but the code is not, so increase the export
image


YoavCodes:
Thanks for the PR. This is already fixed on `main`

PR #337, opened 2026-03-23
fix: Use Bun.spawnSync instead of execSync for better performance and error handling

zhuzipu:
Some computers cannot correctly use tar to extract resources, so Bun.spawnSync is used instead.


YoavCodes:
This isn’t the right fix for the wrapper. The underlying Windows tar issue is already handled on `main` by forcing the built-in system `tar.exe`, and I don’t want the npm/Node entrypoint to depend on Bun, so I’m closing this PR.

PR #350, opened 2026-03-26
fix(core): allow null for BrowserView.ptr type null assignment

AhmedNAgy25:
## What
Fixes a TypeScript type safety issue in `BrowserView` by properly allowing `null` values for the `ptr` property.

- Changed `ptr` from `Pointer!` (non-null assertion) to `Pointer | null = null`
- Removed `null as any` type hack in the `remove()` method

## Why
The `ptr` property represents a native pointer that becomes invalid after `remove()` is called. The previous code set it to `null` but the type didn't reflect this, requiring a type cast hack (`as any`) to work around TypeScript.

This fix:
- Makes the types accurately represent the runtime behavior
- Removes the type hack for cleaner code
- Prevents potential runtime errors from improper null handling


YoavCodes (approved):
Thanks!

PR #352, opened 2026-03-27
Fix Tar in `bunx electrobun init` on windows & ignore node

### LINKED ISSUE #149: Tar extract does not work on Windows using Git Bash
Git Bash on Windows uses GNU tar which mostly works except when it doesn't, in this case it doesn't.

```bash
$ npx electrobun init
Downloading electrobun CLI for your platform...
Error: Failed to download electrobun CLI: Command failed: tar -xzf "C:\Users\henri\AppData\Local\npm-cache\_npx\6f574992512e98b9\node_modules\electrobun\.cache\electrobun-win-x64.tar.gz"
tar (child): Cannot connect to C: resolve failed

gzip: stdin: unexpected end of file
tar: Child returned status 128
tar: Error is not recoverable: exiting now
```

Or in other words it tries to connect to the host `C:` using RSH (???).

`--force-local` could be added to prevent this, or just using Bun itself to extract instead.

Relevant code: https://github.com/blackboardsh/electrobun/blob/7112802023f082366fbef6d46cfcf4bfbe31ac39/package/bin/electrobun.cjs#L97

NullDev:
1. Use the system tar command on windows for `bunx electrobun init` instead of cygwin, gitbash, or gnuwin32
    - Resolves https://github.com/blackboardsh/electrobun/issues/149
2. The init script actually uses nodejs and not bun. Outdated node versions cause errors (e.g. optional chaining)
    - No downside with using bun directly, even in the shebang

<sub><a href="https://discord.com/channels/1190061229914464397/1468277801374453974/1487188026869420154">context</a></sub>

### DIFF
--- a/package/bin/electrobun.cjs
+++ b/package/bin/electrobun.cjs
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env bun
 
 const { execSync, spawn } = require('child_process');
 const { existsSync, mkdirSync, unlinkSync, chmodSync, copyFileSync, createWriteStream } = require('fs');
--- a/package/bin/electrobun.cjs
+++ b/package/bin/electrobun.cjs
@@ -32,18 +32,37 @@ const platform = getPlatform();
 const arch = platform === 'win' ? 'x64' : getArch();
 const binExt = platform === 'win' ? '.exe' : '';
 
+// On Windows, ensure we use the system tar (C:\Windows\System32\tar.exe)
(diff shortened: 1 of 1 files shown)


YoavCodes:
I landed the Windows tar fix directly on `main` as a narrower change that keeps the npm wrapper Node-compatible, so I’m closing this PR.

NullDev:
This still causes problems with conflicting node versions (why the need for node compatibility anyway?) 
So you basically cherry picked 1 of 3 fixes from this PR?

PR #353, opened 2026-03-30
fix: open macOS windows on the active screen

aleclarson:
This fixes macOS window creation always targeting the first display by selecting the screen under the current mouse position and placing the new NSWindow relative to that screen instead, so new windows open on the active monitor rather than always on display 1.


YoavCodes (commented):
This implementation actually breaks a lot of stuff.

You can already create a window on the "active screen" using TypeScript with the existing APIs by using the mouse coordinates to determine the active display and then positioning the window there.

Closing.

PR #354, opened 2026-03-30
Use xcrun for macOS native builds

aleclarson:
This switches the macOS native build steps from plain clang++ to xcrun --sdk macosx clang++ so the build uses Apple’s SDK-aware toolchain even when another clang++ appears first in PATH.


YoavCodes (approved):
Thanks!

Issue #226, opened 2026-02-27
`win.maximize();` makes the window full screen in Windows.


deeivihh:
Calling `win.maximize()` on Windows makes the window go full screen instead of just maximizing it within the desktop. 

### Expected behavior
- Make the window maximized but still show the taskbar.
- Behave like the regular “maximize” button in a typical Windows application.

<img width="1919" height="1079" alt="Image" src="https://github.com/user-attachments/assets/9724c06d-652b-4269-839d-b1ce504c96e7" />
example with windows notepad

### Actual behavior
- The window becomes full screen and covers the taskbar.

<img width="1919" height="1079" alt="Image" src="https://github.com/user-attachments/assets/bad15f2e-50e9-4bd6-8df6-d88f3330c135" />

### Steps to reproduce
- Run an app using electrobun on Windows.
- Obtain the win object in code.
- Call `win.maximize()`;.
- Observe that the window becomes full screen instead of normally maximized.
​
### Environment
OS: Windows 11
electrobun version: 1.14.4

### Additional context 
If this is the intended behavior for `win.maximize()` on Windows, it might be helpful to document it or provide a separate API for “true” fullscreen vs normal maximize.

deeivihh:
any update on this?

YoavCodes:
Fixed in the next stable release.

PR #363, opened 2026-04-05
fix: hiddenInset frameless window on Windows

255doesnotexist:
problem

on windows, \`titleBarStyle: \"hiddenInset\"\` currently uses \`WS_POPUP | WS_THICKFRAME\`. this leaves a ~2-4px caption bar residue at the top. and another choice \`titleBarStyle: \"hidden\"\` drops the DWM shadow, while resize borders behave inconsistently.

root cause

\`WS_POPUP | WS_THICKFRAME\` tells the window manager to draw a thickframe without a caption, but windows still reserves a small non-client area at the top. without handling \`WM_NCCALCSIZE\`, that residue stays. \`WS_POPUP\` also disables the standard DWM frame entirely.

comparison with other frameworks

| framework | windows hiddenInset approach |
|---|---|
| electron | uses \`WS_THICKFRAME | WS_CAPTION\` + custom \`WM_NCCALCSIZE\` handling via \`DwmDefWindowProc\` / chromium ui layer |
| flutter | sets \`WS_POPUP\` and draws its own shadow/resize handles via skia |
| photino | relies on \`DwmExtendFrameIntoClientArea\` with a full-glass effect |
| tauri/wry | delegates to webview2 defaults; custom chrome is typically borderless with os_shadow disabled |

this patch aligns with the electron approach: keep the system frame for shadow and resize, then surgically remove only the caption area in \`WM_NCCALCSIZE\`.

fix

- switch \`hiddenInset\` style to \`WS_CAPTION | WS_THICKFRAME\`. this keeps the system frame (shadow + resize borders) while giving us a hook to remove the caption area.
- handle \`WM_NCCALCSIZE\` in \`WindowProc\`: call \`DefWindowProc\`, then reset \`rgrc[0].top\` to the original window top. this strips the caption bar height without losing the frame.
- send \`SWP_FRAMECHANGED\` after \`SetWindowLongPtr\` so the non-client area recalculates with the new handler active.

verification

- built and tested in a real app (yandu): window is fully frameless, no top residue, four-edge resize works, DWM shadow is present.
- \`bun test src/shared\` passes (48/48).

### DIFF
--- a/package/src/native/shared/callbacks.h
+++ b/package/src/native/shared/callbacks.h
@@ -19,6 +19,9 @@ typedef uint32_t (*HandlePostMessage)(uint32_t webviewId, const char* message);
 typedef const char* (*HandlePostMessageWithReply)(uint32_t webviewId, const char* message);
 typedef void (*AsyncJavascriptCompletionHandler)(const char* messageId, uint32_t webviewId, uint32_t hostWebviewId, const char* responseJSON);
 
+// Window chrome style enum, shared across platforms for titleBarStyle handling
+enum class ChromeStyle : uint32_t { Default, Hidden, HiddenInset };
+
 // Window event callbacks
 typedef void (*WindowCloseHandler)(uint32_t windowId);
 typedef void (*WindowMoveHandler)(uint32_t windowId, double x, double y);
--- a/package/src/native/win/nativeWrapper.cpp
+++ b/package/src/native/win/nativeWrapper.cpp
@@ -4630,6 +4630,7 @@ typedef struct {
     WindowFocusHandler focusHandler;
     WindowBlurHandler blurHandler;
     WindowKeyHandler keyHandler;
+    bool isHiddenInset;
 } WindowData;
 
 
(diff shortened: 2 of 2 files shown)


cyjin-yl:
updated: maximized state now clamps the client area to the monitor work area instead of letting the caption bar show through.

- normal/restored: strips the caption bar completely (behavior unchanged)
- maximized: uses `MonitorFromWindow` + `GetMonitorInfo` to align `rgrc[0].top` with `rcWork.top`, so the webview fills the screen without clipping above the visible area.

verified in a real app: no top residue in normal mode and no caption bar leak when maximized.

cyjin-yl:
<details>
<summary>
remind the implementation refered in this reply is not active now. we use ChromeStyle enum instead. 
</summary>
regarding `isHiddenInset` in `WindowData`:

it's a per-window bool flag used inside the message loop to decide whether `WM_NCCALCSIZE` should strip the caption bar. we don't reuse the `titleBarStyle` string here for two practical reasons:

1. **lifetime safety** - `titleBarStyle` comes in as a transient `const char*` from the zig FFI boundary. storing it as a long-lived pointer in `WindowData` risks a dangling reference once the caller frees the string.
2. **performance** - `WM_NCCALCSIZE` fires repeatedly while the user drags a resize border. a `strcmp` on every message is unnecessary overhead; a bool flag is zero-cost.

during window creation (`createWindowWithFrameAndStyleFromWorker`) we derive the flag from the incoming `titleBarStyle`, then throw away the string. only windows with `isHiddenInset == true` get the custom non-client calculation; all others fall through untouched.
</details>

cyjin-yl:
follow up refactor on top of the fix:

replaced the one off `bool isHiddenInset` in `WindowData` with `ChromeStyle chromeStyle`, an enum class with values `Default`, `Hidden`, and `HiddenInset`. this lines up with the typescript `TitleBarStyle` type and means future chrome variants do not require adding more bools to the struct.

`ChromeStyle` lives in `shared/callbacks.h` because that header is already imported by all three platform native wrappers, so the type is available everywhere without extra includes.

cyjin-yl:
this pr partially resolves #320 and #226.

#320 (reported by @Andreas-Froyland): the hiddenInset residue and broken frame behavior on Windows is addressed by keeping the system DWM frame (WS_CAPTION | WS_THICKFRAME) and stripping only the caption bar in WM_NCCALCSIZE. four edge resize, shadow, and window snapping all work. the maximize/fullscreen confusion mentioned in #226 (reported by @deeivihh) is also fixed: maximized windows now clamp to the monitor work area instead of covering the taskbar.

cc @YoavCodes for review.

YoavCodes (commented):
Thanks!

YoavCodes (approved):

PR #368, opened 2026-04-08
feat(macos): background audio/media playback for WKWebView

COCPORN:
## Summary
- Enables background audio/media playback in WKWebView (`mediaPlaybackRequiresUserAction = NO`, `allowsInlineMediaPlayback = YES`)
- Prevents macOS from suspending the content process when the app window is occluded — without this, audio stops when switching away from the app
- Exposes `backgroundMedia` as a JS-configurable option on BrowserWindow (defaults to `false`, no behavior change for existing apps)

## Test plan
- [x] Verified audio continues playing when app window is behind other windows
- [x] Verified audio continues when switching to another desktop/space
- [x] Verified `backgroundMedia: false` (default) preserves existing behavior
- [x] Running in production for ~2 weeks

### DIFF
--- a/package/src/bun/core/BrowserView.ts
+++ b/package/src/bun/core/BrowserView.ts
@@ -49,6 +49,8 @@ export type BrowserViewOptions<T = undefined> = {
 	startTransparent: boolean;
 	// Set passthrough on the AbstractView at creation (before first paint)
 	startPassthrough: boolean;
+	// Allow background media playback (prevents suspension when app loses focus)
+	backgroundMedia: boolean;
 	// renderer:
 };
 
@@ -106,6 +108,7 @@ export class BrowserView<T extends RPCWithTransport = RPCWithTransport> {
 	sandbox: boolean = false;
 	startTransparent: boolean = false;
 	startPassthrough: boolean = false;
+	backgroundMedia: boolean = false;
 	isRemoved: boolean = false;
 
 	constructor(options: Partial<BrowserViewOptions<T>> = defaultOptions) {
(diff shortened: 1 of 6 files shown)


chrisvariety:
A quick google looks like we probably shouldn't use those private APIs? e.g. https://github.com/ionic-team/cordova-plugin-ionic-webview/issues/286

COCPORN:
Sorry, this is sloppy on my behalf. You bring up a good point. This is (obviously) Claude and I authoring this together, I am not an expert on macOS internals.

Feel free to close this with "won't merge", I'll maintain my own fork of this until I figure out how to do this cleanly in a different way. I personally don't need to be on the AppStore.

What this does is allow for notification sounds to play when long-running tasks complete in Electrobun app, I found that the app would *very quickly* suspend when navigating away from the workspace the Electrobun-app is running on and upon navigating back to the app there would be a stampede of notification-sounds played.

I'm researching alternative routes to make this solid, but it is not high priority for me as in the short term I will just be using this hack, but I might be back with a PR that is better researched and implemented.

COCPORN:
Closing this — we've moved to CEF, so the WKWebView-specific private API workarounds here are no longer needed. If CEF exhibits a similar suspension issue, it'll need a different approach (Chromium switches / CefSettings). Thanks for the review feedback @chrisvariety!

chrisvariety:
@COCPORN awesome, if you land on a CEF solution definitely let me know! I only clicked on this because I will need a similar "please don't suspend me" solution soon, not for audio playing but for a similar long-running background process.

COCPORN:
@chrisvariety At first glance, it looks like this issue simply doesn't surface with CEF — Chromium doesn't seem to do the same occlusion-based suspension that WKWebView does. We haven't done deep testing on it yet, but background audio and processes have been working fine for us without any workarounds. Might be worth just trying your use case on CEF before investing in a solution!

PR #375, opened 2026-04-13
feat: native per-webview proxy support (SOCKS5/HTTP) for all platforms

drksbr:
## Problem

Electrobun currently provides no way to route a specific webview's traffic through a proxy. CEF mode supports `chromiumFlags: { 'proxy-server': '...' }`, but native webview mode (WKWebView/WebView2/WebKitGTK) has no equivalent API.

This blocks use cases like per-session proxy routing, privacy tools, and apps that connect through SOCKS5 tunnels (e.g. SSH, VPN clients).

## Solution

Adds a `proxy` option to `BrowserWindow` and `BrowserView` that configures the native webview's proxy at creation time, using platform-specific APIs:

| Platform | API | Requirement |
|----------|-----|-------------|
| macOS | `WKWebsiteDataStore.proxyConfigurations` via `NWProxyConfig` (Network.framework) | macOS 14+ |  
| Windows | `CoreWebView2EnvironmentOptions.AdditionalBrowserArguments` (`--proxy-server=`, `--host-resolver-rules=`) | WebView2 |
| Linux | `webkit_web_context_set_network_proxy_settings()` with `WEBKIT_NETWORK_PROXY_MODE_CUSTOM` | WebKit2GTK |

Follows the existing `setNextWebviewFlags` pattern to work around Bun FFI parameter count limits.

## API

```typescript
// Route all traffic through a SOCKS5 proxy
  proxy: 'socks5://127.0.0.1:1090',
  // ...other options
});

// Or on a specific BrowserView
const view = new BrowserView({
  proxy: 'http://proxy.example.com:8080',
  // ...other options
});
```

Supported schemes: `socks5://` and `http://`. Passing `null` or omitting the option uses the default (no proxy / system proxy).

## Notes

- macOS implementation requires macOS 14.0+ (guarded with `@available(macOS 14.0, *)`). Earlier versions fall through silently with no proxy applied.
- Does not require CEF — works with the native webview on all platforms.
- No behavior change for existing code (defaults to `null`).


YoavCodes:

Issue #378, opened 2026-04-16
watchIgnore not working


aleclarson:
I'm using Vite in my Electrobun app (for its HMR), so I don't want Electrobun's `dev --watch` process to restart the app when `src` files are changed.

```ts
build: {
  watchIgnore: ["src/**"]
}
```

^ This didn't help. So I consulted GPT-5.4 xhigh and gave it the Electrobun source as context. The issue appears to be that Bun's file-watching API can emit change events for the `"src"` directory (literally the directory, not the file that I actually saved). And of course, `src/**` doesn't match `"src"`, so Electrobun restarts the app.

### Workaround

In case it will help anyone else who runs into this, here is the workaround:

```ts
build: {
  watchIgnore: ["src", "src/**"]
}
```

YoavCodes:
Fixed in the next stable release.

Issue #225, opened 2026-02-27
Generate compile_flags.txt for clangd support


YoavCodes:
Contributors working on the native C++ code (package/src/native/) don't get IDE support        
  (autocomplete, go-to-definition, diagnostics) out of the box because there's no                
  compile_flags.txt or compile_commands.json for clangd to pick up.

  ## Proposal

  Generate a compile_flags.txt file as a side effect of buildNative() in build.ts. The build
  script already resolves all platform-specific flags (including dynamic pkg-config output on
  Linux), so this is just writing them to disk.

  This would give contributors working on the native wrappers automatic clangd support in any
  editor (VS Code, Neovim, Emacs, etc.) without having to manually figure out the right include
  paths and flags for their system.

  ## Notes
  - build.ts remains the source of truth for compile flags — the generated file is purely for IDE
   tooling
  - Especially useful on Linux where include paths vary by distro

YoavCodes:
Fixed in the next stable release.

PR #385, opened 2026-04-22
Improve contributor UX: watchIgnore, compile_flags, and release notes

bigmacfive:
## Summary
- fix `dev --watch` ignore behavior when Bun emits directory-level paths (e.g. `src`) so patterns like `src/**` work as expected
- generate `src/native/compile_flags.txt` during native builds for clangd/editor support across macOS/Windows/Linux
- enable generated GitHub release notes instead of a static repeated release body
- expand `electrobun init` template parsing to support `--template foo`, `-t foo`, and `init <template> <project>`

## Why
- addresses reported watch-ignore mismatch behavior (#378)
- implements requested clangd support path (#225)
- improves release feed readability based on issue feedback (#366)

## Validation
- verified Bun glob behavior (`src/**` does not match `src`, but does match `src/`) and updated matching accordingly
- tested init flows:
  - `electrobun init --template hello-world`
  - `electrobun init sample-a -t hello-world`
  - `electrobun init hello-world sample-b`
- validated updated files parse/build via Bun build API for changed TS entrypoints

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -272,27 +272,8 @@ jobs:
           tag_name: ${{ github.event.inputs.tag || github.ref_name }}
           files: |
             artifacts/**/*.tar.gz
-          generate_release_notes: false
+          generate_release_notes: true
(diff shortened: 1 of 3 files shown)


YoavCodes:

Issue #377, opened 2026-04-16
No type for <electrobun-webview> to import


lordgreg:
Hi there,

_first of all, many thank for this wonderful project, you are doing an awesome job, keep it up!_

I am using typescript and have a part where I have to `toggleHidden()` specific webviews.

```typescript
      for (const [id, ref] of webviewRefs.entries()) {
        // Type casting ref to access toggleHidden assuming the implementation object has it.
        const webviewRef = ref as any;
        if (id === newApp.id) {
          webviewRef.toggleHidden(false);
        } else {
          webviewRef.toggleHidden(true);
        }
      }
```

As you see, I have to set webviewRef to type of any instead of the correct type. Sadly, I havent found any type that would define the `<electrobun-webview>`. If there are none, let me know, I can help with that :)

Regards 👍🏻

YoavCodes:
Fixed in the next stable release.

PR #386, opened 2026-04-22
Add dedicated export paths for webview element types

bigmacfive:
## Summary
- add dedicated package export paths for webview element typing:
  - `electrobun/webview` -> `WebviewTagElement` types
  - `electrobun/wgpu-view` -> WGPU custom element types
- add README usage snippet showing typed `electrobun-webview` access in renderer code

## Why
This lowers friction for users looking for importable custom-element types and makes the intended type import path explicit.

Related: #377

## Validation
- verified `package/package.json` remains valid JSON
- verified export targets exist in the published API layout under `dist/api/browser/`


YoavCodes:

Issue #344, opened 2026-03-25
PATHS.RESOURCES_FOLDER is not exposed


adamsanderson:
The [documentation](https://blackboard.sh/electrobun/docs/apis/paths/) states that: `PATHS.RESOURCES_FOLDER` can be used to get the path to the resources folder, but `RESOURCES_FOLDER` is not exported, only `VIEWS_FOLDER`.

YoavCodes:
Fixed in the next stable release.

PR #387, opened 2026-04-22
Export RESOURCES_FOLDER and VIEWS_FOLDER directly from bun API

bigmacfive:
## Summary
- export `RESOURCES_FOLDER` and `VIEWS_FOLDER` directly from `electrobun/bun` in addition to the existing `PATHS` namespace
- keep `PATHS` behavior unchanged for backward compatibility
- add README usage snippet showing both access styles

## Why
This reduces friction when users expect these path constants to be directly accessible and makes path usage clearer in docs.

Related: #344

## Validation
- verified `src/bun/index.ts` compiles via Bun build API
- confirmed both named exports and `PATHS.*` exports coexist


YoavCodes:

Issue #383, opened 2026-04-19
Linux Use GtkFileChooserNative instead of GtkFileChooserDialog to support native file pickers (KDE, etc.)


mathe00:
### Problem

On Linux, Electrobun uses `gtk_file_chooser_dialog_new()` (`GtkFileChooserDialog`) in all three file dialog code paths in `package/src/native/linux/nativeWrapper.cpp`:

1. `openFileDialog()` (~line 8460) — `Utils.openFileDialog()` API
2. `onRunFileChooser()` (~line 3099) — WebKit `<input type="file">` handler
3. `OnFileDialog()` (~line 1709) — CEF file dialog handler

This means users on KDE Plasma, LXQt, and other non-GTK desktop environments always get the GTK file picker instead of their platform-native one. The `GTK_USE_PORTAL=1` environment variable has no effect because it only works with `GtkFileChooserNative`.

### Proposed Solution

Replace `gtk_file_chooser_dialog_new()` with `gtk_file_chooser_native_new()` (`GtkFileChooserNative`) in all three locations.

`GtkFileChooserNative` (available since GTK 3.20+) automatically delegates to the XDG Desktop Portal (`org.freedesktop.portal.FileChooser`) when available, which uses the desktop environment's native file picker:

- **KDE Plasma** → Dolphin/KDE file dialog (via `xdg-desktop-portal-kde`)
- **GNOME** → GTK file dialog (via `xdg-desktop-portal-gnome`)
- **Flatpak/Sandboxed** → portal dialog automatically
- **Fallback** → standard `GtkFileChooserDialog` (current behavior)

This is the same approach Electron adopted in [PR #19159](https://github.com/electron/electron/pull/19159), and has also been implemented by JavaFX ([openjdk/jfx#2025](https://github.com/openjdk/jfx/pull/2025)) and HandBrake ([HandBrake/HandBrake#4720](https://github.com/HandBrake/HandBrake/pull/4720)).

### Example Change

```cpp
// Before
GtkWidget* dialog = gtk_file_chooser_dialog_new(
    "Open File", NULL, action,
    "_Cancel", GTK_RESPONSE_CANCEL,
    "_Open", GTK_RESPONSE_ACCEPT, NULL);

// After
GtkFileChooserNative* native = gtk_file_chooser_native_new(
    "Open File", NULL, action,
    "_Open", "_Cancel");
GtkWidget* dialog = GTK_WIDGET(native);
```

[...]

PR #389, opened 2026-04-22
Linux: switch file pickers to GtkFileChooserNative

bigmacfive:
## Summary
- replace Linux `GtkFileChooserDialog` usage with `GtkFileChooserNative` in three paths:
  - CEF file dialog handler (`OnFileDialog`)
  - WebKit `run-file-chooser` handler
  - `openFileDialog` utility API
- keep existing filtering and multi-select behavior intact

## Why
`GtkFileChooserNative` allows desktop-native picker integration (including KDE/portal-backed environments) and avoids forcing GTK dialog styling/behavior.

Related: #383

## Validation
- verified all affected call sites now run via `gtk_native_dialog_run(GTK_NATIVE_DIALOG(...))`
- verified file list extraction, cancel flow, and cleanup (`g_object_unref`) remain consistent

### DIFF
--- a/package/src/native/linux/nativeWrapper.cpp
+++ b/package/src/native/linux/nativeWrapper.cpp
@@ -1717,9 +1717,8 @@ public:
         
         printf("CEF Linux: File dialog requested - mode: %d\n", static_cast<int>(mode));
         
-        // Run the file dialog using GTK on the main thread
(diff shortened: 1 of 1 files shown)


YoavCodes:

PR #396, opened 2026-04-23
 feat(win): add native titlebar theme API and defer initial show to avoid caption flicker

adam2am:
## Summary

Adds a Windows native titlebar theme API for `BrowserWindow` and defers the initial HWND show until after the theme is applied.

So when user have a dark theme, we could make it work, and not show a white titlebar.

Flicker-fix fixes a first-paint issue where the window could briefly appear with the default white caption before the requested dark/custom titlebar styling was applied.

## What changed

- added `nativeTitleBar` options to `BrowserWindow`
  - `darkMode`
  - `captionColor`
  - `textColor`
  - `borderColor`
- validate color input in JS before crossing the FFI boundary
- only apply the native titlebar theme on:
  - Windows
  - `titleBarStyle: "default"`
- export no-op `setWindowTitleBarTheme` implementations on macOS/Linux so the FFI surface stays consistent
- stop showing the HWND inside native window creation on Windows
- let the higher-level bridge apply titlebar attributes first, then explicitly show the window

## Why this is safe

Before this change, the window was effectively shown twice on Windows:
- once during native creation
- again from the Bun bridge after setup

This PR removes the earlier native-side show so the theme can be applied before first paint. 
=> visibility is still controlled by the existing higher-level create/show flow.

## Scope

This PR is intentionally limited to the Windows native titlebar theme API and initial-show ordering.

It does not change DirectComposition behavior. The small `dcomp_compositor.h` change only reuses the existing Windows build detection helper needed for titlebar color gating.

## Tests

Added kitchen coverage for:
- creating a window with `nativeTitleBar` options
- rejecting invalid `#RRGGBB` color input before window creation

## Platform behavior

- Windows: applies native titlebar theming when supported
- macOS/Linux: no-op for the new symbol

### DIFF
--- a/kitchen/src/test-framework/executor.ts
+++ b/kitchen/src/test-framework/executor.ts
@@ -121,6 +121,7 @@ export class TestExecutor {
           },
           rpc: options.rpc,
           titleBarStyle: options.titleBarStyle,
+          nativeTitleBar: options.nativeTitleBar,
           trafficLightOffset: options.trafficLightOffset,
           sandbox: options.sandbox || false,
         });
--- a/kitchen/src/test-framework/types.ts
+++ b/kitchen/src/test-framework/types.ts
@@ -57,6 +57,12 @@ export interface WindowOptions {
   y?: number;
   title?: string;
   titleBarStyle?: TitleBarStyle;
+  nativeTitleBar?: {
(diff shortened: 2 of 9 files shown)


YoavCodes:

Issue #357, opened 2026-04-02
Docked DevTools causes webview content to scroll/shift when inspector receives focus


pabloflores465:
## Description

When using docked DevTools (Web Inspector attached to the bottom of the window), clicking/focusing the inspector panel causes the entire webview content to scroll downward, leaving a large empty gray area at the top. The application UI gets pushed to the bottom of the visible area.

When the DevTools are closed or the app regains focus, the content remains displaced and does not recover its original position.

## Steps to Reproduce

1. Create a `BrowserWindow` with `titleBarStyle: "hiddenInset"`
2. Open docked DevTools (Web Inspector attached to the bottom of the window)
3. Click anywhere in the DevTools panel (Elements tab, Console, etc.)
4. Observe: the webview content shifts/scrolls downward significantly

## Expected Behavior

The webview content should remain fixed in place regardless of whether the DevTools panel has focus.

## Actual Behavior

The entire webview content scrolls down when the DevTools inspector receives focus. A large empty gray area appears at the top of the window, and the app UI is pushed to the bottom half. The content does not recover when switching focus back.

## Environment

- **Electrobun version:** 1.16.0
- **macOS version:** Darwin 25.3.0 (macOS Sequoia)
- **Architecture:** arm64 (Apple Silicon)
- **Renderer:** native (WKWebView)

## Analysis

This appears to be a WKWebView-level issue where the internal `NSScrollView` that wraps the webview gets scrolled when the docked inspector panel receives focus. 

We attempted several CSS/JS workarounds from the webview side, none of which had any effect:

- `position: fixed; inset: 0` on `html`/`body`
- `overflow: clip` instead of `overflow: hidden`
- `overscroll-behavior: none`
- Listening for `scroll` events and forcing `scrollTo(0, 0)`
- Using the Visual Viewport API to detect and counteract offset
- Polling with `setInterval` to reset scroll position

None of these worked because the scroll happens at the native `NSScrollView` level, outside the reach of the DOM/JS.

## Possible Fix

The native Zig/ObjC code that creates the WKWebView likely needs to either:
1. Disable scrolling on the enclosing `NSScrollView` (e.g., `scrollView.hasVerticalScroller = false; scrollView.scrollEnabled = false`)
2. Or handle the inspector resize/focus event to reset the scroll view's `contentOffset` to `(0, 0)`

## Workaround

Using DevTools in a **separate/undocked window** avoids the issue entirely. Only docked mode is affected.

## Window Configuration

```typescript
[...]

Issue #190, opened 2026-02-23
DevTools crashes on mouse hover over inspector panel (WKWebView, macOS)


ChrisPeiSC:
## Description

Opening DevTools via `webview.openDevTools()` works, but any mouse hover over the DevTools inspector panel causes an immediate native crash (`EXC_BREAKPOINT / SIGTRAP`). This makes DevTools completely unusable on macOS with the native WKWebView renderer.

This appears to be a continuation of #54 — the underlying WKWebView inspector mouse event handling issue was not fully resolved.

## Steps to Reproduce

1. Create a `BrowserWindow` with a webview
2. Call `mainWindow.webview.openDevTools()` — DevTools opens successfully
3. Move the mouse cursor over the DevTools panel
4. **Crash** — the entire application terminates

`toggleDevTools()` also crashes when closing, and `closeDevTools()` likely has the same issue.

## Environment

- **Electrobun**: 1.13.1 (latest on npm)
- **macOS**: 26.2 (25C56) / Darwin 25.2.0
- **Hardware**: Mac14,9 (M2 Pro), ARM64
- **WebKit**: 21623.1.14.11.9

## Crash Log (key frames)

```
Exception Type:    EXC_BREAKPOINT (SIGTRAP)
Exception Codes:   0x0000000000000001, 0x0000000103cc5a5c

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   bun                                0x103cc5a5c
1   bun                                0x103cb5478
2   bun                                0x1011061b0
3   libsystem_platform.dylib           0x19cfa7744 _sigtramp + 56
4   libNativeWrapper.dylib             0x1051192d8 __31-[WKWebViewImpl toggleDevTools]_block_invoke + 272
5   libdispatch.dylib                  0x19cde0b5c _dispatch_call_block_and_release + 32
...
```

The crash originates in `[WKWebViewImpl toggleDevTools]_block_invoke` within `libNativeWrapper.dylib`. This is the same crash path as #54, where the root cause was `[WKInspectorWKWebView toggleMirrorMode:]` — an unrecognized selector.

## Relation to #54

In #54, @YoavCodes noted:
> "I suspect these code paths have already been refactored away and this bug may just go away in the next version of electrobun."

[...]

Issue #197, opened 2026-02-24
Opening inspector on docked mode create empty space on view


takuyahara:
Opening inspector on docked mode create empty space on view and it persists. Non-docked mode doesn't introduce the issue. The issue is confirmed on [hello-world](https://github.com/blackboardsh/electrobun/tree/main/templates/hello-world) template.

https://github.com/user-attachments/assets/ba2e7bba-ea34-4aa3-98ab-414b76540167

```
macOS: 26.4 Beta (25E5207k)
electrobun: 1.14.4
bun: 1.3.8
Nix: 2.31.1
```

YoavCodes:
Fixed in the next stable release.

PR #398, opened 2026-04-24
fix(macos/wkwebview): detach inspector before show to keep host webview visible

kravetsone:
> [!WARNING]
> **This patch is untested locally** and was authored with significant LLM assistance (Claude). The hypothesis is plausible and matches the observed unified-log signal, but I have not yet rebuilt `libNativeWrapper.dylib` against this change to confirm. Please verify before merging — happy to iterate or close if the hypothesis is wrong.

## Symptom

In a downstream Electrobun app on macOS (Apple Silicon, recent macOS), opening the WebKit inspector via `openDevTools` and then closing the inspector window leaves the host `BrowserWindow` in a broken state: the window remains, the `bun` and host `WebContent` processes are both alive, no `NSException` is raised — but the main `WKWebView` is no longer displayed. On windows configured with `transparent: true` (which is what surfaced this) the user sees an empty/transparent window and assumes the app crashed.

Unified-log signal at the moment the inspector is closed is benign — `WebProcessProxy::canTerminateAuxiliaryProcess: returns true → shutDown` for the inspector's auxiliary `WebContent` process (the inspector UI itself runs as its own WKWebView), no abort, no exception. The host `WebContent` process keeps running, just not painting.

This is **not** the same bug as #190 (hover crash via `toggleMirrorMode:` unrecognized selector). No exceptions are raised here.

Related but distinct: #357 documents another docked-DevTools regression (host webview content scrolls/shifts when the panel opens) — same root area (host view-hierarchy gets disturbed by docked-inspector split-view reparenting), and the workaround listed there (*"use undocked DevTools window"*) is exactly what this PR enforces by default.

## Root-cause hypothesis

`package/src/native/macos/nativeWrapper.mm` opens devtools via:

```objc
id inspector = [self.webView performSelector:@selector(_inspector)];
[inspector performSelector:@selector(show)];
```

Default `_WKInspector show` opens **docked** (attached) — WebKit reparents the host `WKWebView` into a split-view container alongside the inspector view to render them side-by-side in the same `NSWindow`. When the docked panel is dismissed, WebKit's restore path does not always put the host webview back at its original spot in the parent's `subviews` (especially when the host's container is the `NSWindow.contentView` of a transparent, layer-backed, full-size-content window). The host view is then orphaned in the hierarchy.

This matches the symptom precisely: process alive, no exception, no NSWindow gone, but no host content rendered.

The CEF path in this same file already guards against an analogous teardown problem with `RemoteDevToolsWindowDelegate.windowShouldClose:` returning `NO` and calling `orderOut:` — comment: *"Prevent NSWindow from actually closing to avoid CEF teardown crashes."* The WKWebView path has no equivalent guard.

## Fix

Force the inspector into its own `NSWindow` before showing by sending `detach` first. `_WKInspector` exposes `attach`/`detach` selectors (used internally by Safari for its docked/undocked toggle). Calling `detach` before `show` keeps the inspector in a separate window so the host's view hierarchy is never disturbed by reparenting.

```diff
 - (void)openDevTools {
     dispatch_async(dispatch_get_main_queue(), ^{
         if ([self.webView respondsToSelector:@selector(_inspector)]) {
             id inspector = [self.webView performSelector:@selector(_inspector)];
+            if ([inspector respondsToSelector:@selector(detach)]) {
+                [inspector performSelector:@selector(detach)];
+            }
             if ([inspector respondsToSelector:@selector(show)]) {
                 [inspector performSelector:@selector(show)];
             }
         }
     });
 }
```

`respondsToSelector:` guards keep this safe across macOS versions if Apple ever renames the private selector.

## Trade-off

Inspector always opens undocked. Users who previously relied on the docked layout lose that mode. Given the alternative is the host view silently going invisible, undocked-only is the safer default — and matches what other WKWebView-based desktop frameworks (e.g. Tauri) do.

If you want to preserve the docked option behind an opt-in, happy to wire it through the existing `BrowserWindow`/`BrowserView` config surface in a follow-up.

## Repro context

- macOS 15.x, arm64
- `transparent: true`, `titleBarStyle: "hiddenInset"`, `FullSizeContentView` on the host window
- `electrobun` ~1.16.x via npm in the downstream app

## Test plan (for reviewer with build env)

- [ ] Build `libNativeWrapper.dylib` with the patch
- [ ] Open a built sample app, hit `openDevTools` (or its keyboard binding)
- [ ] Confirm inspector appears in its own window
- [ ] Close the inspector window
- [ ] Confirm host window keeps rendering (was previously transparent/empty)
- [ ] Repeat with a non-transparent window to confirm no regression there

### DIFF
--- a/package/src/native/macos/nativeWrapper.mm
+++ b/package/src/native/macos/nativeWrapper.mm
@@ -2990,6 +2990,20 @@ runOpenPanelWithParameters:(WKOpenPanelParameters *)parameters
             // WKWebView doesn't have public DevTools API, but we can use private API if available
             if ([self.webView respondsToSelector:@selector(_inspector)]) {
                 id inspector = [self.webView performSelector:@selector(_inspector)];
+                // Force the inspector into its own NSWindow before showing.
+                // Default `[inspector show]` opens it docked, which reparents
+                // the host WKWebView into a split-view container; closing the
+                // docked panel does not always restore the original parent
+                // and leaves the host view orphaned. On windows configured
+                // with `transparent: true`, the user then sees an empty/
+                // transparent window — the WebContent process is still alive
+                // and there is no NSException, but the main view is no longer
+                // displayed. Detaching first keeps the inspector in a
+                // separate window so the host's view hierarchy is never
+                // disturbed.
+                if ([inspector respondsToSelector:@selector(detach)]) {
+                    [inspector performSelector:@selector(detach)];
+                }
                 if ([inspector respondsToSelector:@selector(show)]) {
                     [inspector performSelector:@selector(show)];
                 }


sahil-shubham:
Related to #197 this too

sroussey:
I just got started with electrobun (day 2) and when i show the devtools all is fine, but then i mouse over the content and back to the devtools and the content hides under the devtools. if i close the devtools then the content is visible, but only where the devtools was before. same if i pop the dev tools into a new window. although it remembers the devtools being outside in its own window, so the next time i don't have the content shift around, which is usable but not ideal.

YoavCodes:

Issue #211, opened 2026-02-25
Feature: Add Hash Support For The Views Protocol


TB516:
Currently, when using the `views://` protocol, trying to open the HTML page with a specific hash set does not work. This would be beneficial to add as using SPA routers are common for desktop apps, and this would let devs launch a new window with a route open by default.

erfanmola:
Up for this, it's a required feature and essential for most frameworks or libraries that support hash routers.

rjschie:
Agree, and I think it's broader than hash support. It'd be nice to treat the `views://` protocol as other URIs per RFC 3986, and account for encoded URI components too. Stripping fragments (`#`) and `?` and beyond them would suffice to help guide users, but since they're valid file name characters, we'd need a way to provide those characters in a file lookup which is the purpose of the URI encoding.

| Before | After Parsed |
|--------|--------|
| `views://mainview/index.html#/some/hash/route` | `views://mainview/index.html` |
| `views://mainview/index.html?key=value` | `views://mainview/index.html` | 

But with a file that contains a legitimate `?` we'd have to encode it so it doesn't get stripped.

| Before | After Parsed |
|--------|--------|
| `views://mainview/legit-file-with-?.html` | `views://mainview/leg-file-with-` (broken!) |
| `views://mainview/legit-file-with-%3F.html` | `views://mainview/leg-file-with-?.html` (fixed!) |

YoavCodes:
Fixed in the next stable release.

PR #405, opened 2026-04-28
Parse URI fragment, query, and percent-encoding in scheme handlers

rjschie:
## Summary

The `views://` scheme handlers passed the raw URL substring (after stripping `views://`) directly to the file system. This treated URI metadata — fragment, query, and percent-encoding — as part of the filename, breaking standard URI semantics.

## What this fixes

Issue: https://github.com/blackboardsh/electrobun/issues/211

Strip `#fragment` and `?query`, then percent-decode the remaining path before resolving to disk. Applied across all five `views://` entry points:

- macOS WKWebView (`normalizeViewsRelativePath`)
- macOS CEF resource handler
- Windows CEF resource handler
- Windows WebView2 `WebResourceRequested` handler
- Windows `handleViewsSchemeRequest`
- Linux WebKitGTK URI scheme handler
- Linux CEF resource handler

## Examples

| URL | Before (file lookup) | After |
|---|---|---|
| `views://mainview/index.html#/settings` | `index.html#/settings` (404) | `index.html` ✓ |
| `views://mainview/data.json?v=2` | `data.json?v=2` (404) | `data.json` ✓ |
| `views://mainview/my%20file.html` | `my%20file.html` (404) | `my file.html` ✓ |
| `views://mainview/café.html` (encoded as `caf%C3%A9.html`) | `caf%C3%A9.html` (404) | `café.html` ✓ |

## Pitfall this resolves

Hash-based SPA routing was unusable. A standard setup —

```js
// router loads views://mainview/index.html, then navigates client-side
location.hash = '#/settings/sub-settings';
```

— worked on first paint, but refresh caused the webview to re-request `views://mainview/index.html#/settings/sub-settings`. The handler tried to open a file named `sub-settings` in the dirs `index.html#/settings` and 404'd, blanking the view. With this fix, refresh resolves to index.html and the router reads the fragment client-side as expected.

The same class of bug affected querystring cache-busting (?v=2) and any file with spaces or non-ASCII characters in the name.

## Test plan

- Load `views://mainview/index.html#/foo` and refresh — page still renders
- Load a view with a querystring — file resolves
- Place a file with a space or unicode char in views and load it
- Existing views without fragment/query still load (regression check)
- All native targets: WKWebView (mac), CEF (mac), WebView2 (win), CEF (win), WebKitGTK (linux), CEF (linux)

### DIFF
--- a/package/src/native/linux/nativeWrapper.cpp
+++ b/package/src/native/linux/nativeWrapper.cpp
@@ -525,6 +525,47 @@ public:
     IMPLEMENT_REFCOUNTING(ElectrobunResponseFilter);
 };
 
+// Percent-decode a URI path component in place per RFC 3986 (UTF-8 byte-wise;
+// invalid escapes are passed through as-is, matching browser behavior).
+static inline void percentDecodeInPlace(std::string& s) {
+    std::string out;
+    out.reserve(s.size());
+    for (size_t i = 0; i < s.size(); ++i) {
+        if (s[i] == '%' && i + 2 < s.size()) {
+            auto hex = [](char c) -> int {
+                if (c >= '0' && c <= '9') return c - '0';
+                if (c >= 'a' && c <= 'f') return 10 + (c - 'a');
+                if (c >= 'A' && c <= 'F') return 10 + (c - 'A');
+                return -1;
+            };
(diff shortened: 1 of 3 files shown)


YoavCodes:

PR #409, opened 2026-05-01
fix(gen-webgpu-ffi): add missing dirname import to template

francip:
## Summary

`scripts/gen-webgpu-ffi.mjs` regenerates `src/bun/webGPU.ts` from a template literal whenever `vendorWGPU()` detects a WGPU version change (`package/build.ts:1146-1149`). The template emits:

```ts
import { existsSync } from "fs";
import { join } from "path";
```

but the generated body uses `dirname(process.execPath)` inside `findWgpuLibraryPath()`. Any consumer that runs after a regen crashes at startup with:

```
ReferenceError: dirname is not defined
    at findWgpuLibraryPath
```

The committed `src/bun/webGPU.ts` is correct (it includes `dirname` in the import); the bug is only triggered on regen, which silently overwrites the committed file with a broken version.

## Fix

One-character template change in `scripts/gen-webgpu-ffi.mjs`: add `dirname` to the path import.

## Test plan
- [ ] `node package/scripts/gen-webgpu-ffi.mjs` regenerates `src/bun/webGPU.ts` with `import { join, dirname } from "path";`
- [ ] App boots without `ReferenceError: dirname is not defined`


francip:
Closing as "won't fix" as per the comment on https://github.com/blackboardsh/electrobun/issues/438

Issue #395, opened 2026-04-23
Windows: titleBarStyle: "hidden" lacks native Win32 title-bar behaviors (Aero Snap, drag-from-maximized, double-click-maximize, system menu) — fixable with ICoreWebView2Settings9::put_IsNonClientRegionSupportEnabled


Spoodies:
On Windows, `titleBarStyle: "hidden"` + the `electrobun-webkit-app-region-drag` CSS class gives you **window dragging only**. Every other native Win32 title-bar behavior that a user expects from a normal window is missing:

1. **Aero Snap** — dragging to screen edges/corners to snap halves/quarters
2. **Drag-from-maximized** — grabbing a maximized window and dragging it down to restore
3. **Double-click-to-maximize** — OS double-click behavior on the drag region
4. **Aero Shake** — shake the window to minimize all others
5. **Alt+Space / right-click system menu** — native window system menu

(Snap Layouts — hover the maximize button for the Win11 layout picker — is out of scope for this issue. It requires `HTMAXBUTTON` from `WM_NCHITTEST` on the parent HWND, which needs either `put_Bounds` shrinking below the caption strip, or the upcoming `CoreWebView2WindowControlsOverlay` API (currently experimental). Tracking separately.)

The current behavior looks like: drag works, but the window feels non-native — it can't snap, can't be restored by dragging from the top of the screen, double-click does nothing, and right-click shows no system menu.

## Environment

- Electrobun: (fill in your version — e.g. `1.14.x`)
- OS: Windows 10 / 11
- WebView2 Runtime: Evergreen (Edge ≥120 is common; setter is gated on this)
- Repro: any `BrowserWindow` with `titleBarStyle: "hidden"` and a `electrobun-webkit-app-region-drag` element

## Reproduction

1. Create a `BrowserWindow` with `titleBarStyle: "hidden"`.
2. Render a title bar `<div class="electrobun-webkit-app-region-drag">` per the docs.
3. On Windows, try: dragging to the left screen edge (no snap preview), double-clicking the title bar (nothing happens), right-clicking the title bar (no system menu), maximizing then dragging down (nothing happens).

Expected: all of the above behave like a normal Win32 window.
Actual: only plain dragging works.

## Root cause

Electrobun's current Windows drag implementation uses a `SetTimer` + `SetWindowPos` loop on the Bun-side main thread. This works for simple drag, but cannot participate in the OS modal move loop that drives Aero Snap, drag-from-maximized, Shake, and the system menu.

[...]

PR #417, opened 2026-05-04
win: enable WebView2 non-client region support and Aero Snap for hiddenInset windows

Spoodies:
Partial fix for #395 (`hiddenInset` only — `hidden` uses `WS_POPUP` so these flags are no-ops there by Win32 design).

  Two changes:
  1. Enable `ICoreWebView2Settings9::put_IsNonClientRegionSupportEnabled(TRUE)` so WebView2 honors CSS `app-region: drag` on its own thread, letting the OS
  run the modal move loop natively. Gracefully no-ops on runtimes older than Edge 120 (SDK 1.0.2420.47).
  2. Add `WS_MAXIMIZEBOX | WS_MINIMIZEBOX | WS_SYSMENU` to the `hiddenInset` window style — without these, Aero Snap, Aero Shake, and the system menu
  silently fail even with #1 in place. `WM_NCCALCSIZE` still collapses the visible caption so no native caption strip is exposed.

  **Test**
  - [ ] Drag title bar to screen top/edges/corners → Aero Snap
  - [ ] Right-click title bar / Alt+Space → system menu
  - [ ] No native caption strip visible

# AI
Hello I wrote this using Claude as it was a problem I was experiencing making my own app using Electrobun. Used Claude cause its out of my depth to solve by myself. I hope this PR is helpful. Thanks :)

### DIFF
--- a/package/src/native/win/nativeWrapper.cpp
+++ b/package/src/native/win/nativeWrapper.cpp
@@ -6056,7 +6056,24 @@ static std::shared_ptr<WebView2View> createWebView2View(uint32_t webviewId,
                             
                             view->setController(ctrl);
                             view->setWebView(webview);
-                            
+
+                            // Enable native Win32 non-client region support so CSS `app-region: drag`
+                            // is honored by WebView2's WndProc on Chromium's thread (which already
+                            // owns input capture), letting the OS run its modal move loop natively.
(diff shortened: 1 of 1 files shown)


YoavCodes:

PR #419, opened 2026-05-06
feature (WIN) / optional custom user agent support for webview2

HarshalKudale:
### Problem 
Current web view implementation does not support custom user agents for webview2, which in my case was required to load TV version of YouTube. 

### Summary
Sets custom User-Agent via the Dev-tools Protocol (Emulation.setUserAgentOverride) using `ICoreWebView2::CallDevToolsProtocolMethod` When a non-empty `user-agent` is provided, the web-view uses the Dev-tools level override so that the effective User-Agent header matches the provided string and User-Agent Client Hints (Sec-CH-UA* headers) are not sent.

Tested with my own YouTube app (not published) using local linking.


YoavCodes:

PR #422, opened 2026-05-07
fix: exit dev --watch process when app window closes

mekwall:
## Problem

When running `electrobun dev --watch`, closing the app window leaves the dev server process running indefinitely. The terminal appears to hang, requiring a manual `Ctrl+C` to terminate.

## Root Cause

The `onExit` callback in `runDevWatch()` only nulls the `appHandle` reference but never triggers the parent process to exit. Meanwhile `await new Promise(() => {})` keeps the event loop alive forever.

## Fix

The `onExit` callback is kept minimal (`appHandle = null` only). The actual exit logic lives in `appHandle.exited.then()` where the real exit code is available and rebuild state can be checked:

```ts
// Rebuild path — guards against watcher-driven restarts
appHandle.exited.then((code) => {
    if (!isBuilding && !shuttingDown) {
        cleanup();
        process.exit(code);
    }
}).catch(() => {});

// Initial launch path
appHandle.exited.then((code) => {
    if (!shuttingDown) {
        cleanup();
        process.exit(code);
    }
}).catch(() => {});
```

Key design decisions:
- **Exit code propagation** — uses the actual exit code from the app process instead of hardcoded `0`
- **Rebuild-safe** — the `!isBuilding` guard in the rebuild path ensures `appHandle.kill()` during rebuilds does NOT trigger cleanup or exit
- **Atomic cleanup+exit** — `cleanup()` and `process.exit()` happen together, so `shuttingDown` (set by cleanup) cannot block the exit
- **Unhandled rejection safety** — `.catch(() => {})` on the `exited` promise

### DIFF
--- a/package/src/cli/index.ts
+++ b/package/src/cli/index.ts
@@ -4632,6 +4632,15 @@ usageDescriptions : ""}${urlTypes ? "\n" + urlTypes : ""}${documentTypes ?
 						appHandle = null;
 					},
 				});
+
+				if (appHandle) {
+					appHandle.exited.then((code) => {
+						if (!isBuilding && !shuttingDown) {
+							cleanup();
+							process.exit(code);
+						}
+					}).catch(() => {});
+				}
 			} catch (error) {
 				console.error("[electrobun dev --watch] Build failed:", error);
 				console.log("[electrobun dev --watch] Waiting for file changes...");
@@ -4687,6 +4696,15 @@ usageDescriptions : ""}${urlTypes ? "\n" + urlTypes : ""}${documentTypes ?
 					appHandle = null;
 				},
 			});
+
+			if (appHandle) {
+				appHandle.exited.then((code) => {
+					if (!shuttingDown) {
+						cleanup();
+						process.exit(code);
+					}
+				}).catch(() => {});
+			}
 		} catch (error) {
 			console.error("[electrobun dev --watch] Initial build failed:", error);
 			console.log("[electrobun dev --watch] Waiting for file changes...");


mekwall:
Addressed both Copilot review comments in 636817ab:

1. **Rebuild path** — both `runApp()` calls (initial launch + rebuild) now share the same exit handling
2. **Exit code propagation** — uses `appHandle.exited.then(code => process.exit(code))` instead of hardcoded `process.exit(0)`

mekwall:
Addressed all review feedback in `9ba9690a`:

1. **Rebuild path** — both `runApp()` calls now have identical exit handling via `appHandle.exited.then()`
2. **Exit code propagation** — `process.exit(code)` uses the actual exit code from the app process
3. **Rebuild-safe** — rebuild path uses `!isBuilding && !shuttingDown` guard so `appHandle.kill()` during rebuilds does NOT trigger cleanup or exit. Only user-initiated closes (where `isBuilding` is false) cause the dev server to exit.
4. **Atomic cleanup+exit** — `cleanup()` and `process.exit(code)` happen together in `exited.then`, so `shuttingDown` cannot block the exit (it is set by `cleanup()` but the `process.exit()` call follows immediately)
5. **Unhandled rejection** — added `.catch(() => {})` to `exited.then()` chains

The key change: `onExit` is now a simple `appHandle = null` with no side effects. All cleanup/exit logic lives in `exited.then()` where the actual exit code is available and the `isBuilding` flag can be checked.

YoavCodes:

Issue #328, opened 2026-03-19
Feature Request: Window hide() and isVisible() API


xdmden:
## Summary

`BrowserWindow` currently supports `minimize()`, `unminimize()`, `focus()`, and initial `hidden: true` creation — but there's no way to hide an already-visible window without minimizing it, and no way to query visibility.

## Use case

My app uses a "hotkey window" pattern — a floating panel that appears on a global shortcut and disappears instantly on blur or Escape. This requires hide/show without minimize animations or dock indicators. This is a common pattern for launcher-style apps (Raycast, Alfred, Spotlight) and utility tools.

The existing `minimize()` is not a substitute because it triggers a dock animation and changes the window's minimized state, which is semantically different from simply hiding it.

## Suggested API

```typescript
const win = new BrowserWindow({ hidden: true, ... });

win.show();       // Already exists (calls focus internally)
win.hide();       // Hide without minimizing — the missing piece
win.isVisible();  // Query whether the window is currently visible
```

## Platform implementation notes

- **macOS:** `[window orderOut:nil]` to hide, `[window makeKeyAndOrderFront:nil]` to show
- **Windows:** `ShowWindow(hwnd, SW_HIDE)` / `ShowWindow(hwnd, SW_SHOW)`
- **Linux:** `gtk_widget_hide()` / `gtk_widget_show()`

YoavCodes:
Fixed in the next stable release.

PR #436, opened 2026-05-14
fix(macos): bind default key equivalents for App/Window menu roles

mattd233:
## Problem

Role-based menu items don't bind the standard macOS key equivalents on their own. `{ role: "hide" }`, `{ role: "hideOthers" }`, `{ role: "quit" }`, `{ role: "minimize" }`, `{ role: "close" }` all resolve to the correct selectors via `getMenuRoleToSelectorMap`, but the `if (!accelerator) { … }` ladder in `createMenuFromConfig` only assigns default `keyEquivalent`s for Edit-menu roles. So the menu items ship with `keyEquivalent = @""` and the system shortcuts (⌘H, ⌘⌥H, ⌘Q, ⌘M, ⌘W) don't fire.

Related to but narrower than #328 (which requests a JS-side `BrowserWindow.hide()` / `isVisible()` API). This PR doesn't add any JS surface; it just makes role-based menu items behave like the rest of macOS.

## Fix

Extend the existing role → default-keyEquivalent ladder with the five App/Window roles that have well-known system shortcuts. `showAll`, `zoom`, `bringAllToFront`, `toggleFullScreen`, etc. are intentionally left without defaults because there's no single Apple convention or the binding has shifted (e.g. fullscreen).

## Test plan

- [x] App menu `{ role: "hide" }` — ⌘H hides the app.
- [x] App menu `{ role: "hideOthers" }` — ⌘⌥H hides other apps.
- [x] App menu `{ role: "quit" }` — ⌘Q quits.
- [x] Window menu `{ role: "close" }` — ⌘W closes the focused window.
- [x] Existing Edit-menu roles (`undo`/`redo`/`cut`/`copy`/`paste`/`selectAll`/etc.) — no regression.

### DIFF
--- a/package/src/native/macos/nativeWrapper.mm
+++ b/package/src/native/macos/nativeWrapper.mm
@@ -1170,6 +1170,21 @@ NSMenu *createMenuFromConfig(NSArray *menuConfig, StatusItemTarget *target) {
                     } else if ([role isEqualToString:@"selectAll"]) {
                         menuItem.keyEquivalent = @"a";
                         menuItem.keyEquivalentModifierMask = NSEventModifierFlagCommand;
+                    } else if ([role isEqualToString:@"hide"]) {
+                        menuItem.keyEquivalent = @"h";
+                        menuItem.keyEquivalentModifierMask = NSEventModifierFlagCommand;
+                    } else if ([role isEqualToString:@"hideOthers"]) {
+                        menuItem.keyEquivalent = @"h";
+                        menuItem.keyEquivalentModifierMask = NSEventModifierFlagCommand | NSEventModifierFlagOption;
+                    } else if ([role isEqualToString:@"quit"]) {
+                        menuItem.keyEquivalent = @"q";
+                        menuItem.keyEquivalentModifierMask = NSEventModifierFlagCommand;
+                    } else if ([role isEqualToString:@"minimize"]) {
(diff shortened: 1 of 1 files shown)


package/src/native/macos/nativeWrapper.mm
```diff
@@ -1170,6 +1170,21 @@ static void parseMenuAccelerator(NSString *accelerator,
...
+                    } else if ([role isEqualToString:@"hide"]) {
+                        menuItem.keyEquivalent = @"h";
+                        menuItem.keyEquivalentModifierMask = NSEventModifierFlagCommand;
+                    } else if ([role isEqualToString:@"hideOthers"]) {
+                        menuItem.keyEquivalent = @"h";
+                        menuItem.keyEquivalentModifierMask = NSEventModifierFlagCommand | NSEventModifierFlagOption;
+                    } else if ([role isEqualToString:@"quit"]) {
+                        menuItem.keyEquivalent = @"q";
+                        menuItem.keyEquivalentModifierMask = NSEventModifierFlagCommand;
+                    } else if ([role isEqualToString:@"minimize"]) {
```
mattd233, line 1182: This actually already works for some reason after adding the Window menu using `ApplicationMenu.setApplicationMenu` ,I haven't verified why that works, it might be falling back on some system default but is added for completeness/correctness.

YoavCodes:

PR #444, opened 2026-05-20
upgrade react-tailwind-vite template with react v19 tailwind v4 vite v8

modejota:
Basically, as the title says, update the template to use more modern versions of react, vite and tailwindcss


YoavCodes:

PR #451, opened 2026-05-27
fix: prevent WKWebsiteDataStore crash on older macOS versions

edengilbertus:
## Summary
- Added `-mmacosx-version-min=10.13` to `libNativeWrapper.dylib` compilation to ensure `@available` macros aren't overly optimized.
- Added a `respondsToSelector:` safeguard around `dataStoreForIdentifier:` to prevent an unrecognized selector crash on macOS < 14.

## Problem
Currently, the app instantly crashes on macOS 12 (and likely 13) with `+[WKWebsiteDataStore dataStoreForIdentifier:]: unrecognized selector sent to class`.

## Solution
This adds the required minimum deployment target to the compilation phase and adds a runtime method check to safely fallback to `defaultDataStore` on older macOS versions.

### DIFF
--- a/package/build.ts
+++ b/package/build.ts
@@ -1836,8 +1836,8 @@ async function buildNative() {
 		const wgpuIncludeFlag = existsSync(wgpuIncludeDir)
 			? `-I${wgpuIncludeDir}`
 			: "";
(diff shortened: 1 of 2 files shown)


YoavCodes:

These older macOS versions are also outside of our support matrix

PR #454, opened 2026-06-01
feat: add will-close event to intercept X button before window is destroyed

### LINKED ISSUE #455: Feature Request: Intercept window close button (X) before window is destroyed
Problem:
The existing close event fires from windowWillClose: — after AppKit has already committed to destroying the window. Common patterns like "unsaved changes?" dialogs or saving state before quit are impossible at this point because the window is already gone.

Expected behavior:
A way to intercept the X button click while the window is still alive, optionally cancel the close, do async work, then close programmatically.

Suggested API:
```
win.on('will-close', (event) => {
    event.response = { allow: false };
    saveDocument().then(() => win.close());
});
```

Notes:
- AppKit's windowShouldClose: is the correct veto point — it fires before windowWillClose: and can return NO to cancel
- Handlers would need to be synchronous (same constraint as Electron's close event) — async work is done after returning false, then win.close() is called programmatically
- win.close() calls [window close] directly which bypasses windowShouldClose:, so no loop
- Follows the same pattern Electron uses for its close event

A POC implementation is linked in the PR: https://github.com/blackboardsh/electrobun/pull/454

akleventis:
> **POC / suggested implementation** — not expecting a merge, opening to express the idea.

Closes #455

## Problem

The existing `close` event fires from `windowWillClose:` — after AppKit has already committed to destroying the window. Common patterns like "unsaved changes?" dialogs or saving state before quit are not possible at this point because the window is already gone.

## Approach

Hooks `windowShouldClose:` (AppKit's veto point, fires *before* `windowWillClose:`) to emit a `will-close` event while the window is still alive. If a handler sets `event.response = { allow: false }`, the close is cancelled. Otherwise the window closes normally — no behavior change for existing code.

## Usage

```typescript
win.on('will-close', (event) => {
    event.response = { allow: false }; // must be synchronous
    saveDocument().then(() => win.close());
});
```

**Handlers must be synchronous.** To do async work (show a dialog, save a file), set `allow: false` immediately to block the close, then call `win.close()` programmatically once done. `win.close()` calls `[window close]` directly, which bypasses `windowShouldClose:` — no loop. Same pattern Electron uses for its `close` event.

## Changes

- `callbacks.h` — `WindowShouldCloseHandler` typedef
- `nativeWrapper.mm` — `windowShouldClose:` calls handler and returns `NO`; returns `YES` when no handler is registered (no behavior change for existing code)
- `main.zig` — trampoline + `should_close_handler` in `WindowState`; trampoline only wired when handler is non-null
- `native.ts` — `windowShouldCloseCallback` JSCallback, passes as arg to `createWindow`
- `windowEvents.ts` — `willClose` event with `{ allow: boolean }` response type
- `zig-sdk/electrobun.zig` — `WindowShouldCloseHandler` type + `should_close` in `WindowCallbacks`

### DIFF
--- a/package/src/bun/events/windowEvents.ts
+++ b/package/src/bun/events/windowEvents.ts
@@ -13,6 +13,8 @@ type KeyData = { id: number; keyCode: number; modifiers: number; isRepeat: boole
 
 export default {
 	close: (data: IdData) => new ElectrobunEvent<IdData, {}>("close", data),
+	willClose: (data: IdData) =>
+		new ElectrobunEvent<IdData, { allow: boolean }>("will-close", data),
 	resize: (data: ResizeData) =>
 		new ElectrobunEvent<ResizeData, {}>("resize", data),
 	move: (data: MoveData) => new ElectrobunEvent<MoveData, {}>("move", data),
--- a/package/src/bun/proc/native.ts
+++ b/package/src/bun/proc/native.ts
@@ -154,6 +154,7 @@ const core = (() => {
 					FFIType.function,
 					FFIType.function,
 					FFIType.function,
+					FFIType.function,
 				],
 				returns: FFIType.u32,
 			},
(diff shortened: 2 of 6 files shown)


YoavCodes:

PR #494, opened 2026-07-13
add macOS tray length option

nasedkinpv:
Adds `TrayOptions.length?: number | "square"` for macOS status items.

- `number`: fixed width in points
- `"square"`: system square width
- omitted: existing variable sizing

The value is preserved when the tray is recreated.

Tested on macOS 26.5:

- `18`: `18×22`, including after hide/show
- `"square"`: `22×22`
- unit tests, typecheck, docs check, Zig/core and native builds


YoavCodes:

PR #495, opened 2026-07-14
fix: isolate sandboxed webview events

ShiroKSH:
## Summary

- bind sandbox event dispatch to the native sender id
- serialize special event detail as data while preserving JSON and string fallback behavior
- add focused regressions for spoofed ids and encoded host events

## Root cause

Sandboxed content could select another webview through its payload id, and special event detail was embedded directly in a JavaScript expression evaluated by the host webview.

## Checks

- bun test src/shared/naming.test.ts src/sdks/bun/proc/eventBridge.test.ts
- bunx tsc --noEmit for eventBridge source and test
- runtime JavaScript regression for encoded special event detail


YoavCodes:

PR #506, opened 2026-07-26
fix(win): read LOCALAPPDATA as wide string for WebView2 userDataFolder

### LINKED ISSUE #505: [Windows] Non-ASCII %LOCALAPPDATA% (e.g. CJK username) breaks WebView2 user data folder — app launches but renders nothing
## Environment

- Windows 11, Korean locale, user profile `C:\Users\김종환` (Hangul username)
- Electrobun 1.18.1, system webview (WebView2), packaged app built via `electrobun build --env=stable`

## Symptom

The app installs fine (self-extractor, Start Menu shortcut, WebView2 runtime all OK) and the launcher/bun processes start, but no window content ever renders. WebView2 shows an error dialog whose path contains mojibake: the username renders as `◆◆◆◆ø` instead of `김종환`, and `...\WebView2\Partitions\default\EBWebView` is never created. Creating the same directory manually with `mkdir` succeeds, so it is not a permissions/disk issue.

## Root cause

In `package/src/native/win/nativeWrapper.cpp` (`createWebView2View`):

```cpp
char* localAppData = getenv("LOCALAPPDATA");
...
int wideSize = MultiByteToWideChar(CP_UTF8, 0, userDataPath.c_str(), -1, nullptr, 0);
```

`getenv()` returns the value in the **active code page** (CP949 on Korean Windows), but the conversion assumes **CP_UTF8**. CP949 bytes for Hangul are invalid UTF-8, so the path is mangled before being handed to `SHCreateDirectoryExW` / `CreateCoreWebView2EnvironmentWithOptions`, and environment creation fails. Any user whose Windows profile path contains non-ASCII characters (Korean, Japanese, Chinese, Cyrillic, accented Latin, ...) hits this with the default system-webview config.

## Other occurrences of the same pattern

- Same `getenv("LOCALAPPDATA")` in the CEF cache path setup (also passed to `CreateDirectoryA`).
- `buildAndEnsurePartitionCachePath()` — same `getenv`, plus a naive `std::wstring(cachePath.begin(), cachePath.end())` widening that is only correct for ASCII.

This is the same ANSI-vs-wide family as #335 (menus/file dialogs).

## Fix

Read the variable via the wide API and convert to UTF-8 before building the path, so the existing UTF-8 → wide conversion round-trips. PR incoming for the launch-blocking WebView2 site (minimal diff, reuses the file's existing `WStringToString()` helper); the CEF sites need a couple of additional narrow-API changes and are listed here so they can be tackled with the same approach.

JongDeug:
Fixes #505.

## Problem

`createWebView2View` builds the WebView2 user data folder from `getenv("LOCALAPPDATA")`, which returns **active-code-page** bytes (CP949 on Korean Windows), then converts the assembled path with `MultiByteToWideChar(CP_UTF8, ...)`. For any non-ASCII Windows username the bytes are invalid UTF-8, the path is mangled (`김종환` → `◆◆◆◆ø`), folder creation fails, and `CreateCoreWebView2EnvironmentWithOptions` never succeeds — the app launches but renders nothing. This hits the default (system webview) config for every user with a CJK/Cyrillic/accented profile path.

## Fix

Read the variable with `_wgetenv(L"LOCALAPPDATA")` and convert wide → UTF-8 with the file's existing `WStringToString()` helper. The downstream UTF-8 → wide conversion then round-trips correctly. 5 lines changed, no new helpers, no behavior change for ASCII paths.

The two CEF-path occurrences of the same pattern are intentionally left out to keep this reviewable — they need accompanying narrow-API changes (`CreateDirectoryA`, a naive `char`→`wchar_t` widen) and are documented in #505.

## Verification

Failure reproduced on a Korean-locale Windows 11 machine with a Hangul username (mojibake path visible in the WebView2 error dialog; `WebView2\` directory never created; manual `mkdir` of the same path succeeds, ruling out permissions). The patched path logic mirrors the `WStringToString` usage already present elsewhere in this file.

### DIFF
--- a/package/src/native/win/nativeWrapper.cpp
+++ b/package/src/native/win/nativeWrapper.cpp
@@ -6746,9 +6746,12 @@ static std::shared_ptr<WebView2View> createWebView2View(uint32_t webviewId,
             // Create user data folder path based on partition
             // Build path with identifier/channel structure (consistent with CLI and updater)
             std::wstring userDataFolder;
-            char* localAppData = getenv("LOCALAPPDATA");
+            // Read via the wide CRT API: getenv() returns active-code-page bytes
+            // (e.g. CP949 on Korean Windows), which corrupt non-ASCII profile
+            // paths when converted as UTF-8 below.
+            wchar_t* localAppData = _wgetenv(L"LOCALAPPDATA");
             if (localAppData) {
-                std::string userDataPath = buildAppDataPath(localAppData, g_electrobunIdentifier, g_electrobunChannel, "WebView2", '\\');
+                std::string userDataPath = buildAppDataPath(WStringToString(localAppData), g_electrobunIdentifier, g_electrobunChannel, "WebView2", '\\');
 
                 // Handle partition-specific storage
                 if (!partitionStr.empty()) {


YoavCodes:

PR #477, opened 2026-06-20
fix(linux): stop CEF window move/resize crash (Xlib threading + ConfigureNotify storm)

srsholmes:
## Problem

On Linux, dragging a CEF `BrowserWindow` — **resizing from a corner** or **moving it between monitors** — crashes the app with `Signal 11` (sometimes `SIGABRT`), with seemingly **random backtraces**, often deep inside the JS engine's GC (`JSC::MarkingConstraint::execute`, etc.). Reproducible in a fresh `electrobun init` CEF app on a multi-monitor X11 setup. Affects 1.16.0 through current `main` (1.18.4-beta.6).

This appears to be the cause behind several open reports:
- Closes #426 (Crash on linux, mostly on resize) — same `Signal 11` + random JSC-GC backtrace.
- Related #281 (GLXBadWindow / Signal 11 on Wayland/CEF).
- Related #145 (window move handling — `startWindowMove`).
- Related #371 (Linux resize issues).

## Root cause

**Xlib is driven from two threads with no `XInitThreads()`.**

- `ElectrobunClient::OnPaint` runs on the **CEF UI thread** and calls `XCreateImage` / `XPutImage` / `XFlush` on the OSR buffer.
- `process_x11_events` runs on the **main / GTK thread** and calls `XPending` / `XNextEvent`.
- Both use the **same `Display*` connection** (`EnableOSR(x11win->window, x11win->display, …)`).

Xlib is not thread-safe unless `XInitThreads()` is called before the first Xlib call. Without it, the two threads racing on one `Display` during a move/resize **event storm** corrupt Xlib's internal request buffer; the corruption surfaces later as a crash in an unrelated thread (usually the JS GC) — which is exactly why the reported backtraces look random and unrelated to windowing.

I confirmed the crashing thread was inside `process_x11_events` via coredump symbol resolution, and that `XInitThreads()` alone makes resize stop crashing.

## Fix

1. **`XInitThreads()`** as the first call in `initializeGTK()` (before `gtk_init()` and any `XOpenDisplay()`), so Xlib serializes cross-thread access.
2. **Coalesce the `ConfigureNotify` storm**: a drag emits many per second. Collapse the pending burst to the latest geometry with `XCheckTypedWindowEvent`, then fire the JS move/resize callbacks + CEF `WasResized()`/OSR re-render **at most once per drain**, and only when the position (move) or size (resize) actually changed. A pure move no longer triggers an OSR repaint storm. This also removes a redundant `resizeCallback` firing on position-only changes.

[...]

PR #511, opened 2026-08-02
fix(linux): scale nested CEF webviews on HiDPI displays

### LINKED ISSUE #510: Linux CEF: nested electrobun-webview bounds are not scaled on HiDPI displays
## Description

On Linux with bundled CEF, a nested `<electrobun-webview>` is positioned and sized incorrectly when the display scale is greater than 100%. At 200% scaling, the child appears at half the expected physical offset and occupies half the expected width and height. The host DOM element itself has the correct layout.

This prevents the documented “super iframe” behavior: the native child does not fill its CSS Grid/Flexbox cell even though the custom element does.

## Reproduction

1. Run an Electrobun app with Linux `bundleCEF: true` on a HiDPI display (reproduced at 200%).
2. Put `<electrobun-webview>` inside a normal CSS Grid or Flexbox workspace next to a sidebar.
3. Size it with ordinary CSS (`width: 100%; height: 100%`).
4. Compare the element bounding box with the native CEF child window.

The DOM `getBoundingClientRect()` is correct in CSS/DIP pixels, but the native child is visibly smaller and offset toward the top-left.

## Root cause

The preload overlay contract correctly sends CSS/DIP coordinates. On Linux, CEF windowed children are X11 windows, but the initial `CefWindowInfo::SetAsChild` bounds and later `XMoveResizeWindow` calls use those DIP values directly as physical X11 pixels. Mask rectangles follow the same incorrect coordinate path.

Full-size BrowserViews are different: their resize path already receives raw physical dimensions from X11 `ConfigureNotify` events and must not be scaled again.

## Expected behavior

At every supported display scale, a nested `<electrobun-webview>` should exactly match its host element. Fixed child bounds and masks should be converted from DIPs to physical pixels exactly once at the Linux native boundary, using the CEF scale factor for the parent window monitor. Full-size BrowserViews should retain their current physical resize path.

## Environment

- Electrobun: current `main` (`1.18.4-beta.12`)
- OS: Ubuntu 26.04 LTS
- Session: GNOME Wayland; Electrobun CEF uses its X11/XWayland path
- Display scale: 200%

## Duplicate check

I searched open and closed issues and pull requests for Linux CEF, HiDPI, nested webview, scaling, bounds, `devicePixelRatio`, and X11 resize terms. The nearest items address different paths:

- #302 synchronizes a full-window CEF view after a tiling window manager resizes its parent.
- #477 addresses Xlib thread safety and `ConfigureNotify` storms.
- #324 is the analogous Windows WebView2 DPI issue.

None converts Linux nested child-view bounds from DIPs to X11 pixels.

Ferrb9579:
## Summary

- resolve the CEF display scale for the X11 parent window using physical monitor coordinates
- convert nested CEF child bounds from public DIPs to X11 physical pixels during initial creation and subsequent resizes
- scale visual/input mask rectangles through the same native boundary
- preserve the existing raw-X11 sizing path for full-size BrowserViews to avoid double scaling
- add edge-based fractional-scale conversion tests and source-contract coverage

Fixes #510.

## Why the conversion is native

`overlaySync` already defines the cross-platform contract correctly: DOM geometry stays in CSS/DIP pixels. Windows performs its DIP-to-native conversion in the native wrapper. Doing the Linux conversion beside `SetAsChild`, `XMoveResizeWindow`, and XShape handling keeps that contract consistent and prevents extensions/apps from applying renderer-specific DPR workarounds.

The helper scales both rectangle edges instead of independently rounding origin and size, so adjacent elements remain adjacent at fractional scale factors.

## Testing

- `bun test src/shared src/sdks/bun src/config src/preload` — 169 passed
- native C++ conversion test compiled and passed with `g++ -std=c++17`
- manually verified in Butler on Ubuntu/GNOME at 200% display scaling: a grid/flex-hosted `<electrobun-webview>` now fills the complete workspace to the right of the sidebar and below the title bar

## Related work

This is intentionally separate from #302 (full-window tiling-WM resize synchronization) and #477 (Xlib thread safety and configure-event coalescing). This patch targets fixed/nested child geometry and leaves full-size X11 dimensions unscaled.

### DIFF
--- a/package/dash.config.ts
+++ b/package/dash.config.ts
@@ -26,10 +26,11 @@ export default {
 		"push:stable": "hutch check:release && node scripts/push-version.js stable",
 		"build:push:artifacts": "node scripts/build-and-upload-artifacts.js",
 		"test:dialog-paths-native": "hutch scripts/test-dialog-paths-native.js",
+		"test:linux-dpi-native": "hutch scripts/test-linux-dpi-native.js",
 		"test:windows-ui-native": "hutch scripts/test-windows-ui-native.js",
 		"test:windows-ui-native-integration":
 			"hutch scripts/test-windows-ui-native.js --require-native-wrapper",
-		"test:unit": "node scripts/run-cottontail-test.js src/shared src/sdks/bun src/config src/preload && hutch test:dialog-paths-native && hutch test:webview2-permissions && hutch test:windows-ui-native",
+		"test:unit": "node scripts/run-cottontail-test.js src/shared src/sdks/bun src/config src/preload && hutch test:dialog-paths-native && hutch test:linux-dpi-native && hutch test:webview2-permissions && hutch test:windows-ui-native",
 		"test:linux-native-dialog":
 			"node scripts/run-cottontail-test.js src/shared/linux-native-file-dialog.test.ts && scripts/test-linux-native-file-dialog.sh",
 		"test:cef-debug": "scripts/test-cef-remote-debugging.ts",
--- a/package/scripts/test-linux-dpi-native.js
+++ b/package/scripts/test-linux-dpi-native.js
@@ -1,0 +1,67 @@
+import { existsSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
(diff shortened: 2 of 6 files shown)


YoavCodes:

PR #553, opened 2026-09-10
fix(templates): add electrobun.config.ts to tsconfig.json include array

### LINKED ISSUE #541: In version 2.0.0.beta13, there is a type error in electrobun.config.ts
There is a minor type error in the latest test version. This issue will be fixed by modifying the template file, as the "include" field in the tsconfig.json file does not list electrobun.config.ts, causing ```import type { ElectrobunConfig } from "electrobun";``` to fail because electrobun cannot be found.
Solution:
``` json
{
    "extends": "./.hutch/devkit/tsconfig.json",
    "compilerOptions": {
        "target": "ESNext",
        "useDefineForClassFields": true,
        "module": "ESNext",
        "resolveJsonModule": true,
		"allowJs": true,
        "isolatedModules": true,
        "moduleDetection": "force",
        "noEmit": true,
        "strict": true,
        "noUnusedLocals": true,
        "noUnusedParameters": true,
		"noFallthroughCasesInSwitch": true,
        "moduleResolution": "bundler"
    },
    "include": ["src"],
    "exclude": ["node_modules", "dist", "build", "../../package/dist"]
}
```
Replace with
```
	"include": ["src","./electrobun.config.ts"],
```
By the way, are there any plans to support TypeScript 6.0 or even 7.x?

MHAlikhani:
## Summary

Adds `./electrobun.config.ts` to the `include` array in every template `tsconfig.json`, so the root-level config file falls inside the TypeScript program and module resolution works correctly.

## Problem

Template `tsconfig.json` files either only included `"src"`, or had no `include` array at all. Since `electrobun.config.ts` lives at the project root — outside `src/` — it was excluded from the compilation scope, and TypeScript reported:

> Cannot find module 'electrobun' or its corresponding type declarations.

for this import:

```ts
```

## Solution

Include the config file explicitly in every template:

- Templates that already had `include: ["src"]` → extended to `["src", "./electrobun.config.ts"]`
- Templates with no `include` field at all → added `"include": ["./electrobun.config.ts"]`

No other fields, files, or templates were touched.

## Testing

- Confirmed every template `tsconfig.json` now references `./electrobun.config.ts`
- Verified each edited file is still valid JSON (no trailing commas or syntax errors)
- `git diff` reviewed to ensure only the `include` arrays changed

Closes #541


MHAlikhani:
@YoavCodes  just following up on #553 in case it got buried.

The change is limited to the 31 template `tsconfig.json` files and fixes the project scope for `electrobun.config.ts`. It’s ready for review whenever you get a chance.

